PR #2, opened 2014-04-26
DuckDNS

gkranis:
Added simple support for Duckdns www.duckdns.org ,seems to work fine.
DuckDNS itself allows you to pass empty ip= argument to autodetect
 but I haven't added support for that.
Please review as I'm not familiar with perl
(Was mostly copy-paste but you never know :) )

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -4071,6 +4071,8 @@ EoEXAMPLE
 ######################################################################
 ## nic_duckdns_update
 ## by George Kranis (copypasta from nic_dtdns_update)
+## http://www.duckdns.org/update?domains=mydomain1,mydomain2&token=xxxx-xxx-xx-x&ip=x.x.x.x
+## response contains OK or KO
 ######################################################################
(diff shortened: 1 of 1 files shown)


wimpunk:
I merged your reqeuest but I'm missing some configuration example.  See https://sourceforge.net/p/ddclient/wiki/protocols/#your-favorite-provider-here for more info about what I need.

caviedes55:
Hi @wimpunk, do you need something more? I´ve noticed all this points are completed, or not?

    it's a free dynamic DNS provider.
    if the provider support SSL please add it to the readme.ssl file.
    if you provide a nice example in the sample-etc_ddclient.conf file.
    if you don't forget to edit the README.md
    if your patch applies nicely
    and if you provide a link to the API.
Cheers

wimpunk:
I should check the ddclient tree but it looks like I was missing some configuration example in `sample-etc_ddclient.conf`

PR #7, opened 2014-09-07
Added support for Cloudflare and multi domain support for namecheap

roberthawdon:
As featured at: https://support.cloudflare.com/hc/en-us/articles/200168816-Does-CloudFlare-work-with-Dynamic-DNS-Can-I-update-my-DNS-records-remotely-

Namecheap improvements: https://robertianhawdon.me.uk/2010/09/03/making-ddclient-work-with-multiple-domains-on-namecheap/


Bugsbane:
> I would like to request that "use JSON::Any;" be made conditional on selecting the cloudflare provider.

I second the motion. I'm not using cloudflare, and currently starting ddclient just gives me the following:

> Mar 08 12:07:07 hal systemd[1]: ddclient.service: control process exited, code=exited status=2
> Mar 08 12:07:07 hal systemd[1]: Failed to start Dynamic DNS Update Client.
> Mar 08 12:07:07 hal systemd[1]: Unit ddclient.service entered failed state.
> Mar 08 12:07:07 hal systemd[1]: ddclient.service failed.
> Mar 08 12:07:07 hal ddclient[14499]: Can't locate JSON/Any.pm in @INC (you may need to install the JSON::Any module) (@INC contains: /usr/lib/perl5/site_perl /usr/...nt line 28.
> Mar 08 12:07:07 hal ddclient[14499]: BEGIN failed--compilation aborted at /usr/bin/ddclient line 28.

I assume this is because of @abelbeck 's issue mentioned in the comment above. Commenting out line 28 lets ddclient start properly again.

wimpunk:
@Bugsbane and @abelbeck, I think you will like commit 98060e2 where I made the use of json optional.  Is there a cloudflare user out there who can test my fix?

Bugsbane:
\o/

zkrx:
Looks like Namecheap multi-domain support has been reverted somehow.
Why? Is there a better way to achieve this that I'm unaware of?

I can't have multiple domain names pointing to the same machine.

Also see https://github.com/linuxserver/docker-ddclient/issues/15 (not sure if it is the same ddclient).

zkrx:
So there must be a regression somewhere. I can't remember the exact behavior off-hand but updating multiple tld's with the same IP on Namecheap does not work. Only the first tld is updated IIRC.

icolwell:
Hi, I also experienced an issue with updating multiple namecheap domains from the same ddclient config. I was using ddclient 3.9.0. It looks like someone opened a [sourceforge bug report](https://sourceforge.net/p/ddclient/bugs/89/) back in August pointing to [r204](https://sourceforge.net/p/ddclient/code/204/) as the issue.

PR #11, opened 2014-12-29
odd-fw-patch-squashed

Tatsuya-Nonogaki:
This is the squashed commits of former odd-fw-patch and its addition, which will add functions to;
* un-zeropad IP address returned by some odd routers,
* make $ip UNDEF if get_ip('fw') returned a local IP address.


wimpunk:
Fixed the issue with cloudfare a while ago. Your patch can be applied.

PR #15, opened 2015-01-19
Renaming Changelog -> ChangeLog.old.

nelsonjr:
On systems that cannot/possibly not care about file case sensitivity (FAT,
NTFS, HFS+) having files with same name but different casing confuses Git, as
it thinks one of the files is deleted.


wimpunk:
It's a good idea to change it but it would be better to move Changelog to Changelog.old and only use ChangeLog as euh changelog.  There are a few references to that file on sf.net so I would like to keep that file.

nelsonjr:
I did not touch the ChangeLog file. Is sf.net case sensitive?

wimpunk:
Oops, sorry, I didn't looked into the patch. Just renaming Changelog to Changelog.old would be good enough.

nelsonjr:
Just to make sure... I noticed that ChangeLog (capital) uses a per SVN revision style of what's new (like svn 520) while the Changelog (lower) is more like a what's in when you declare a new release (e.g. 3.8.2). Are you not putting what's new per version in a file anymore?

nelsonjr:
Done.

wimpunk:
The ChangeLog was generated basesd on subversion commits which previously wasn't used. As changelog is the old one, I simple renamed it to changelog.old en kept the new one. This has been done in b92a4e4.

PR #17, opened 2015-02-15
Fix for CloudFlare default config and update interval

shikasta-net:


wimpunk:
Your patch does much more than what you says it does.  Make your patch as minimal as possible and split it in two pull requests: one adding the config line for cloudfare and one which reads the refresh interval.

PR #20, opened 2015-03-08
Add support for Yandex.Mail for Domain DNS service (domain.yandex.com)

Akimkin:
Yandex's DNS management API (https://api.yandex.com/domain/doc/about.xml) uses POST method and utilizes additional HTTP header for client authentication.
Implementing support for these required some changes in the _geturl_ routine.
Please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Dynamic DNS services currently supported include:
     nsupdate    - See nsupdate(1) and ddns-confgen(8) for details
     CloudFlare  - See https://www.cloudflare.com/ for details
     Google      - See http://www.google.com/domains for details
+    Yandex      - See https://domain.yandex.com/ for details
 
 DDclient now supports many of cable/dsl broadband routers. 
 
--- a/ddclient
+++ b/ddclient
@@ -451,6 +451,9 @@ my %variables = (
 	'googledomains-common-defaults'       => {
 		'server'	      => setv(T_FQDNP,  1, 0, 1, 'domains.google.com', undef),
     },
+    'yandex-common-defaults'       => {
+        'server'              => setv(T_FQDNP,  1, 0, 1, 'pddimp.yandex.ru', undef),
+    },
 );
 my %services = (
     'dyndns1' => {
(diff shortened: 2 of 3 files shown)


wimpunk:
I need to manually review the patch.  It's pretty big and it doesn't apply clean.

Akimkin:
The main steps my patch does are:
1. Connect to HTTP API server
2. Make an API request to retrieve domain records
3. Parse the reply (which is in a JSON format) in order to find the ID of required record.
4. Make another API request to update a record with a new value.

Each reply contains "success" field which indicates whether the request was successfully processed by server, so this field is checked after each request.

The main problems were following:
- Yandex's API uses POST request method instead of GET.
- By some reason their API requires clients to pass an auth key via custom HTTP header ("PddToken") in order to authenticate.

These things were not supported by ddclient, so I needed to modify the *geturl* routine in order to support them. I added two more args to it: data (string containing POST request payload) and headerref (a reference to a hash containing any additonal HTTP headers for the request). If the data is non-empty, then request type is set to POST, otherwise it defaults to GET. The custom headers are put to the request after the "Host: ...." line. Also, some additional headers are set for POST requests.

If the approach I used is acceptable, I can update the patch in order to match latest revision.

DaveSophoServices:
@Akimkin Is the code still valid for Yandex? Do you know if there is a test site we can use to verify the API is still current?

Akimkin:
@DaveSophoServices some time ago Yandex introduced Yandex.Connect service that is intended to replace old one called Yandex PDD (which this patch was made for). Unfortunately, official Yandex.Connect API docs do not contain anything about managing DNS. I found [a not-too-old thread on one Russian Q&A site](https://toster.ru/q/544909) where it's said that Connect services have no such APIs yet (Yandex support says that "they're working on it") but PDD API used in my patch should work for domains that weren't migrated to Connect. New domains can only be added via Connect.

DaveSophoServices:
For reference, here's the documentation from Yandex:
https://tech.yandex.com/domain/doc/concepts/api-dns-docpage/
Looks like it hasn't been degraded. (https://tech.yandex.com/domain/)

DaveSophoServices:
@Akimkin Are you in a position to test the merged code? I've brought in your code with a few minor changes, allowing for the work that's occurred between then and now

Akimkin:
@DaveSophoServices sorry for some delay, I've missed your message. Code logic seems to work well and update the record even though I've migrated to Yandex.Connect. There are couple problems:

- Major one: source needs `use JSON::Any;` declaration, otherwise it fails with `perhaps you forgot to load "JSON::Any"?` error. At least on my Perl version (5.30).
- Minor one: `JSON::Any` module is deprecated, so maybe migrate it to `JSON::MaybeXS` which is (according to docs) a more up-to-date alternative.

PR #21, opened 2015-03-17
merge in service-common-defaults for cloudflare

Harry-Xue:
Otherwise won't read login and password from conf file.


Harry-Xue:
Already pending as request #19.

PR #22, opened 2015-03-20
loopia.se changed the "Current Address:" output string to "Current IP Ad...

reddyr:
loopia.se changed the "Current Address:" output string to "Current IP Address:"


wimpunk:
Thanks for the request!

PR #28, opened 2015-10-23
freedns: send IP address

ony:
http://freedns.afraid.org/ automatically detects external IP. To
override that IP it is possible to supply address=IP as additional
parameter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Dynamic DNS services currently supported include:
     nsupdate    - See nsupdate(1) and ddns-confgen(8) for details
     CloudFlare  - See https://www.cloudflare.com/ for details
     Google      - See http://www.google.com/domains for details
-    Duckdns      - See https://duckdns.org/ for details
+    Duckdns     - See https://duckdns.org/ for details
     woima.fi    - See https://woima.fi/ for details
 
 DDclient now supports many of cable/dsl broadband routers. 
@@ -42,8 +42,9 @@ REQUIREMENTS:
 - one or more accounts from one of the dynamic DNS services
 
 - Perl 5.014 or later
-  (you need the IO::Socket::SSL perl library for ssl-support
-  and JSON::Any perl library for JSON support)
+  (you need the IO::Socket::SSL perl library for ssl-support,
+  JSON::Any perl library for JSON support and
+  IO::Socket:INET6 perl library for ipv6-support)
 
 - Linux or probably any common Unix system
 
--- a/ddclient
+++ b/ddclient
@@ -1944,6 +1944,26 @@ EOM
 	import JSON::Any;
 }
 ######################################################################
+## urladd
+######################################################################
+sub urladd {
+    my $url      = shift || '';
+    debug("url    = %s", $url);
+    unless ($url =~ /(\?[^#]*)?(#.*)?$/) {
+	    failed("invalid url: %s", $url);
+	    return undef;
+	}
+	return $url if (@_ == 0);
+	my ($base, $query, $fragment) = ($`, $1, $2);
+	if (!$query) {
+		$query = '?' . shift;
+	}
+	foreach my $param (@_) {
+		$query .= '&' . $param;
+	}
+	return $base . $query . $fragment;
+}
+######################################################################
 ## geturl
 ######################################################################
 sub geturl {
@@ -3806,7 +3826,7 @@ sub nic_freedns_update {
 	    $config{$h}{'status'} = 'good'; 
 	    success("update not necessary %s: good: IP address already set to %s", $h, $ip); 
 	} else {
-	    my $reply = geturl(opt('proxy'), $freedns_hosts{$h}->[2]);
+	    my $reply = geturl(opt('proxy'), urladd($freedns_hosts{$h}->[2], "address=$ip"));
 	    if (!defined($reply) || !$reply) {
 	        failed("updating %s: Could not connect to %s.", $h, $freedns_hosts{$h}->[2]);
 		last;
--- a/ddclient
+++ b/ddclient
@@ -345,12 +345,12 @@ my %variables = (
 	'retry'               => setv(T_BOOL,  0, 0, 0, 0,                    undef),
 	'force'               => setv(T_BOOL,  0, 0, 0, 0,                    undef),
 	'ssl'                 => setv(T_BOOL,  0, 0, 0, 0,                    undef),
-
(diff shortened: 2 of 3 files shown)


wimpunk:
Could you add some info about how to use it to the readme?  And could you some info about how it works to this pull request?

ony:
@wimpunk, I'm not sure what you mean. There is no change in usage of `ddclient`. `$ip` is detected as usual on one machine I `use=cmd` on another `use=if, if=tun7`.
Before this change detected IP had no effect on the IP with which `freedns` server updated a record. Moreover for a correct work `ddclient` detected IP had to match with IP autodetected on server side. If there were no match that caused a useless updates with error response.

ony:
@wimpunk you can find description on http://freedns.afraid.org/dynamic/
> TIP: You can override automatic IP detection host by appending &address=127.0.0.10 to any update URL (IPv6 addresses are also allowed)

When you use direct URL `http://freedns.afraid.org/dynamic/update.php?xxxxxxx` it always updates with an IP requests comes from. But if you add `address` param it will explicitly request to associate specified IP with domain name.

Without this pull-request `ddclient` supply no IP information with update. But infrastructure of script expect that IP changes will be detected before any update attempt. Thus there is 4 ways to detect IP `ip` (static), `if` (via interface name), `web` (via some external http service) and `cmd` (output from locally executed command). With this change `ddclient` will use that information during updating `freends`.

To add parameter to URL I've added `urladd` function. It takes URL as a first parameter and interpret rest of parameters as a list of query chunks which should be added to url. These will be separated by `&` and will be inserted after `?` (start of query) but before `#` (start of fragment).. Of course it is possible to use some URI parser library, but it looks like script follows a bit different ideology.

wimpunk:
@ony, I'm sorry it took a while to look into this.  I had to read your explication a few times before I understood what you mean.  I didn't realized the original code didn't used the IP fetched by `ddclient`.
As far as I understand your code you are removing this auto-detect feature.  As this changes the way `ddclient` work, I believe it is better to keep the current way of working and add a parameter which make freedns use the IP detected by `ddclient`.  You could add something like a `noautoip` parameter which disables to auto-detection by freedns.

ony:
@wimpunk, do you understand that `ddclient` initially independently of FreeDNS have feature of auto-detect (`ip`, `if`, `web` - similar to what FreeDNS does). I'd say that prerequirement for `ddclient` to be able to detect IP address before doing any DNS updates.
FreeDNS have autodetect because it expect also client that simply use `curl` without any `ddclient`.
Currently these autodetection features are in conflict for me because they detect different IP addresses.
I see confusion if configuration will look like:
1. Here is how to detect IP I want to update
2. Use dyndns2 server to update that record with detected IP
3. Use freedns server to update that record but ignore detected IP and use what freedns provides.

Note that having this `noatuoip` breaks persistent state of `ddclient`  (`ddclient.cache`) which expect that if FreeDNS successfully were updated it will store IP detected by `ddclient` and not one detected by FreeDNS server.
Also FreeDNS server do not report IP if it wasn't changed which makes it impossible to fix `ddclient.cache` in that case.

Sorry. I'm out of ideas how to make it clear for you either because lack of my English or because lack of something else. I don't insist on accepting this change.
Since this is open-source we can maintain our own forks.
So you have choice:
- close it
- accept it as is
- accept it and then add own commits that does what you think is better

wimpunk:
Since I'm maintaining `ddclient` for almost 10 years now, I think I understand how `ddclient` works.  What you found is actually a bug in `ddclient`: the provider should be updated to the IP `ddclient` detects and apparently it doesn't work that way.
What I trying to say is there should be an option to make it possible to use `ddclient` the old way.  If there are some users out there who uses `ddclient` with the IP detected by freedns, they should be able to keep using the service that way.
So if you don't want to add an option to do so, I will but it will take a little longer before your patch will be added to the official one.

ony:
> What you found is actually a bug in ddclient: the provider should be updated to the IP ddclient detects and apparently it doesn't work that way.
@wimpunk yes, that's exactly what I think also.
But do you agree that in configuration it should fit naturally. I can see it fitting normally if it will look like:
```
use=freedns
server=freedns.afraid.org, ...
```
But then it will conflict with possibility to use multiple DNS services in a single config.

PIXELSTURM:
Hi!

First, ony is right. Sorry of. ;)

Second, currently i have the prob, the ddclient grab the wrong account datas at afraid.org. I have 1 account for 2 domains. One for IPv4 and one for IPv6. Different domain names. If i try to update the IPv6 domain, the IPv4 data is taken. At the end has the IPv4 an AAAA address...

Greets.

dkerr64:
ddclient does not support IPv6 for freedns/afraid.org.  The IPv6 support it has is incomplete.  For more complete IPv6 you will need the code changes proposed in pull request #40 which does support freedns and fixes the problem documented here that the IP address is not sent properly.

#40 has not been merged yet and I do not know if there are plans to do so. 

Thx

ony:
> What I trying to say is there should be an option to make it possible to use `ddclient` the old way. If there are some users out there who uses `ddclient` with the IP detected by freedns, they should be able to keep using the service that way.

@wimpunk could you provide a correct scenario for "old way"?
As far as I can see current behavior:
* If IP detected by `ddclient` itself matches with one detected by `afraid.org` then everythin ok (with and without this patch).
* If IP detected by `ddclient` doesn't match with one detected by `afraid.org` then we start to spam `afraid.org` because they are awlays different and require synchronization (`ddclient` becomes dummy `curl`). With this patch IP will be set according to settings in `ddclient`, will check prev value and thus wouldn't spam `afraid.org`.

I'm pretty sure that second case will cause user to give up on `ddclient`.

Since you maintain this project for 10 years you may know some aspects that I don't know (ex. maybe you expect usecase when user didn't specified the way IP is being detected). And maybe you have an idea how it should behave.
P.S. I hope your idea wouldn't sound like "you jsut need an option that will enable this patch". Because that's exactly the way how code starts to fall apart.

PIXELSTURM:
Hi!

Finally i fixed my freedns ipv6 problem with a ≈ 600 Byte python script. It works with ipv4 and ipv6.

For ddclient is my tip: Keep it simple.
So, grabbing the account data from freedns is buggy and too much. The user should pick the "direct update url" via webgui from freedns and put that data in the config. url+newip is the way.

Have fun. Greets.

SuperSandro2000:
I am going to close this because I assume that https://github.com/ddclient/ddclient/commit/566c3c3d5e5b323560f593d6d6109a3f948bf924 fixed all outstanding bugs. If this is not the case please reopen the discussion.

PR #29, opened 2015-10-27
add ipv6 support 

geraldhansen:
added ipv6 support and tested with dyndns service from http://myonlineportal.net


wimpunk:
Your patch looks good and I never heard about the provider but they seem to be a good free one.  I will be happy to try them after merging your pull request but could you also update the README file?

geraldhansen:
what should I update in the README file - the provider I mentioned?
But the provider is using the standard dyndns2 protocol or you like to have some more information about ipv6 in the README file?

wimpunk:
New version of this pull is known as #36.

PR #30, opened 2015-11-28
FIxing warnings on FreeDNS update NOP

hank:
When FreeDNS was updating, I was noticing that it would send back an HTTP 200 with the text "Address [IP] has not changed", but ddclient was issuing a failed status for this and logging a ton of text to syslog.  I changed the function a tiny bit to allow this to be a success status.  I also reformatted the code to all spaces instead of mixed spaces and tabs, but you can take that or leave it - I don't really care, it was just a total whitespace disaster before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 ===============================================================================
-# DDCLIENT v3.8.2
+# DDCLIENT v3.8.3
 
 ddclient is a Perl client used to update dynamic DNS entries for accounts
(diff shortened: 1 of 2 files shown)


wimpunk:
You changed much more than only the part you said.  Can you please create a pull request with only your freedns patch?

hank:
#31 supersedes this pull request.

PR #31, opened 2015-11-29
Amended: Fixing warnings on FreeDNS update NOP

hank:
Amended pull request of #30.  

When FreeDNS was updating, I was noticing that it would send back an HTTP 200 with the text "Address [IP] has not changed", but ddclient was issuing a failed status for this and logging a ton of text to syslog. I changed the function a tiny bit to allow this to be a success status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 ===============================================================================
-# DDCLIENT v3.8.2
+# DDCLIENT v3.8.3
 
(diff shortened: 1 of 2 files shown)


hank:
I've been running this for the last day, and it looks pretty good:
```
Nov 29 14:18:33 xxx ddclient[1949]: SUCCESS:  updating xxx.xxx.xxx: good: IP address x.x.x.x has not changed
```

PR #32, opened 2016-01-10
Added MyDNS.JP

epgdatacapbon:
Add support for MyDNS.JP
http://www.mydns.jp/


epgdatacapbon:
Updated script will be submitted.

PR #36, opened 2016-01-18
add ipv6 support 

geraldhansen:
Hi I rewrite my old ipv6 support,
its now an extra option and works also with use=if
I also add some info to the README.md
Please let me know if you have some further questions or if I miss something.

Best Gerald


jwillmer:
How does IPv6 support work? Does it only work if I have an IPv6 and no IPv4 or do I specify that I am only interested in IPv6 with the IPv6-flag?

niclashoyer:
`ipv6=yes` has no effect for me :disappointed:

dkerr64:
> `ipv6=yes` has no effect for me 😞

You could try this version of ddclient and let us know if it works...
https://github.com/astlinux-project/ddclient-curl

Thanks

PR #37, opened 2016-01-26
Add support for telling `nsupdate` to use TCP instead of UDP

beamerblvd:
By default, `nsupdate` uses UDP unless the update size is too large to fit in a UDP datagram, in which case it automatically switches to TCP. This change adds a `tcp` configuration option to the `nsupdate` protocol so that the user can force `nsupdate` to use TCP.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -437,6 +437,7 @@ my %variables = (
     'nsupdate-common-defaults' => {
 	'ttl'                 => setv(T_NUMBER, 0, 1, 0, 600,                 undef),
 	'zone'                => setv(T_STRING, 1, 1, 1, '',                  undef),
+	'tcp'                 => setv(T_BOOL,   0, 1, 1, 0,                   undef),
     },
 	'cloudflare-common-defaults'       => {
 		'server'	      => setv(T_FQDNP,  1, 0, 1, 'www.cloudflare.com', undef),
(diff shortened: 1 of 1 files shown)


beamerblvd:
As posted in https://sourceforge.net/p/ddclient/discussion/399428/thread/06dc6d2d/.

wimpunk:
This looks pretty similar to
http://sourceforge.net/p/ddclient/discussion/399428/thread/06dc6d2d/ ... :-)

beamerblvd:
Yes, I created that SF discussion when I thought ddclient was still using Subversion and I needed to submit a patch. When I discovered that ddclient now uses GitHub pull requests, I created this pull request. Much easier than patches. :-)

wimpunk:
Your patch looks pretty good.  I'll try to add it asap.

beamerblvd:
I was pleased with how easy it was to add the config option and get it to work. Just two lines of code and four lines of documentation.

beamerblvd:
I see a lot of major changes have been made since 3.8.3 was released, so I'm guessing that the next version will be 3.9.0 or 4.0.0. Would you be amenable to creating a branch at some stable point after 3.8.3, cherry-picking this commit to that branch (after merging it), and releasing a 3.8.4 very soon?

wimpunk:
If you have any suggestion on what I need to add to the protocols wiki, feel free to send me the info. :-)

PR #39, opened 2016-02-12
Allow public ip from any interface

MichaelHarder:
This change will allow a public IP address from any interface to be used if use=if and if=''. It also removes the default setting of if=ppp0 so that by default any public ip address will be used.


wimpunk:
I'm sorry but I don't understand why you need this setting.  What' s wrong with using 'use=if and if=ethx'?  Changing the default behavior of ddclient is something which I try to avoid because there people out there relying on this default behavior.

PR #40, opened 2016-03-18
Comprehensive IPv6 framework with support for freedns.

dkerr64:
Updates here are based on work I have done for the AstLinux project (AstLinux.org)

Adding comprehensive framework for IPv6 support.  IPv6 and IPv4 exist side-by-side
The same network interface may have both IPv4 and IPv6.  Therefore to support both
we add the following options.

 usev6 = [ ip, if, cmd, web ]
 ipv6 = [ set an IPv6 address (usev6=ip) ]
 webv6 = [ url of a web server that will return your IPv6 address (usev6=web)]
 webv6-skip = [ string to skip to before looking for IPv6 address]
 cmdv6 = [ system command to execute that will return an IPv6 address (usev6=cmd)]
 cmdv6-skip = [ string to skip to before looking for IPv6 address ]

There is no ifv6. It is assumed that if=interface will be the same for both IPv4
and IPv6

You can mix methods for IPv4 and IPv6 for the same interface.  For example
 use=web
 usev6=if
 if=eth0
 web=http://ipdetect.dnspark.com/
Would be a valid method for a system behind an IPv4 NAT connection, IPv6 would be
obtained from the eth0 interface (globaly routable address) and the public IPv4
from external web site.

Support for updating AAAA (IPv6) records is added for freedns.afraid.org.  The
existing IPv6 nsupdate method was updated to account for new design.

Optional support is added for cURL as some embedded systems do not have perl
NET6 and SSL packages installed.  Set with option
 curl=yes

Submitted code is ready for code review and comment.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -409,8 +409,6 @@ my %variables = (
 	'cmd-skip'            => setv(T_STRING,0, 0, 1, '',                   undef),
 	'cmdv6'               => setv(T_PROG,  0, 0, 1, '',                   undef),
 	'cmdv6-skip'          => setv(T_STRING,0, 0, 1, '',                   undef),
-
-
 	'ip'                  => setv(T_IP,     0, 1, 0, undef,               undef),
 	'ipv6'                => setv(T_IPV6,   0, 1, 0, undef,               undef),
 	'wtime'               => setv(T_DELAY,  0, 1, 1, 0,                   interval('30s')),
@@ -2288,35 +2286,35 @@ sub get_ip {
     $arg = '' unless $arg;
 
     if ($use eq 'ip') {
-		$ip  = opt('ip', $h);
-		$arg = 'ip';
+	$ip  = opt('ip', $h);
+	$arg = 'ip';
 
     } elsif ($use eq 'if') {
-		$skip  = opt('if-skip', $h)  || '';
-		$reply = `ifconfig $arg 2> /dev/null`;
-		$reply = `ip addr list dev $arg 2> /dev/null` if $?;
-		$reply = '' if $?;
+	$skip  = opt('if-skip', $h)  || '';
+	$reply = `ifconfig $arg 2> /dev/null`;
+	$reply = `ip addr list dev $arg 2> /dev/null` if $?;
+	$reply = '' if $?;
 
     } elsif ($use eq 'cmd') {
-		if ($arg) {
-			$skip  = opt('cmd-skip', $h)  || '';
-			$reply = `$arg`;
-			$reply = '' if $?;
-		}
+	if ($arg) {
+	    $skip  = opt('cmd-skip', $h)  || '';
+	    $reply = `$arg`;
+	    $reply = '' if $?;
+	}
 
     } elsif ($use eq 'web') {
-		$url  = opt('web', $h)       || '';
-		$skip = opt('web-skip', $h)  || '';
+	$url  = opt('web', $h)       || '';
+	$skip = opt('web-skip', $h)  || '';
 
-		if (exists $builtinweb{$url}) {
-			$skip = $builtinweb{$url}->{'skip'} unless $skip;
-			$url  = $builtinweb{$url}->{'url'};
-		}
-		$arg = $url;
+	if (exists $builtinweb{$url}) {
+	    $skip = $builtinweb{$url}->{'skip'} unless $skip;
+	    $url  = $builtinweb{$url}->{'url'};
+	}	    
+	$arg = $url;
 
-		if ($url) {
-			# when using a web server to find public IPv4 address we should force use of IPv4
-			$reply = geturl(opt('proxy', $h), $url, undef, undef, undef, undef, undef, '4') || '';
+	if ($url) {
+	    # when using a web server to find public IPv4 address we should force use of IPv4
+	    $reply = geturl(opt('proxy', $h), $url, undef, undef, undef, undef, undef, '4') || '';
         }
 
     } elsif (($use eq 'cisco')) {
@@ -2372,8 +2370,7 @@ sub get_ip {
         }
         if (!defined $reply) {
         $reply = '';
-	}
-	
+    }
     if ($skip) {
         $skip  =~ s/ /\\s/is;
         $reply =~ s/^.*?${skip}//is;
@@ -2384,7 +2381,7 @@ sub get_ip {
 	    $ip = filter_local($ip) if opt('fw-banlocal', $h);
     }
     if (($use ne 'ip') && (define($ip,'') eq '0.0.0.0')) {
-		$ip = undef;
+	$ip = undef;
     }
 
     debug("get_ip: using %s, %s reports %s", $use, $arg, define($ip, "<undefined>"));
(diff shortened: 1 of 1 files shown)


ddclient
```diff
@@ -2075,6 +2176,37 @@ sub geturl {
...
+	    verbose("SENDING:", "%s", '${protocol}://${server}/${url}');
+        } else
+        {
+            $0 = sprintf("%s - curl sending to %s", $program, '${protocol}://${server}/${url}');
+            my $timeout = opt('timeout');
+
+            $reply = `/usr/bin/curl -si0 --user "${login}:${password}" --user-agent "${program}/${version}" \\
+              --connect-timeout $timeout --max-time $timeout $curlopt \\
+              --url "${protocol}://${server}/${url}" 2>/dev/null`;
+
```
abelbeck: Excellent idea to optionally use ```curl``` as it is ubiquitous, performs either IPv4/IPv6 and SSL all in one call.

But, may I suggest you use "HERE Documents" to minimize unwanted special-character processing:
````
	$reply = <<`USE_CURL`;
/usr/bin/curl -si0 --user '${login}:${password}' --user-agent '${program}/${version}' \\
	  --connect-timeout $timeout --max-time $timeout $curlopt \\
	  --url '${protocol}://${server}/${url}' 2>/dev/null
USE_CURL
```
Also use singe-quotes around the string arguments, better to allow ${password} special characters and does not effect the ${foo} perl replacement.

dkerr64:
Lonnie, thanks for the comments.  I am not familiar with the HERE Documents method but based on your comments it does look like being a good idea.  

Thanks

dkerr64:
Further testing has revealed that the Perl built-in NET6 support is not a reliable method.  While it works fine if you are on a system that is IPv6 connected, it is failing in some cases if you try and use it on a system that is only IPv4.  ie, running "ddclient -query" fails half way through testing webv6 sites if you are on a system that is not currently connected IPv6.  It simply crashes with no diag info on the syswrite command ..
`my $result = syswrite $sd, $rq;`

One argument would of course be that you simply shouldn't do that... use INET6 on a system not IPv6 enabled.  But a simpler approach might just be to enforce use of Curl when IPv6 is requested.

David

andig:
Whats the status of IPv6 support? I have a small raspi with a global IPv6 address that I'd like to tie to my domain?

dkerr64:
@andig  I am using the IPv6 support I added with freedns.afraid.org and it is working fine.  I use it on two systems... my gateway, which obtains the IPv6 address from the ethernet interface and I use it on an internal server (behind IPv4 NAT) which obtains IPv6 addresses for several internal servers and updates them to freedns (if you want the outside world to get to those servers then you also need to add appropriate firewall rules to the gateway).

However it works only with the freedns service.  I have not and don't plan to try and make it work with any other service.  I leave that to someone else.

@wimpunk has not commented on the changes I submitted and has not merged into the master.  I would very much like it merged but that is not up to me.  I would highly recommend a thorough code review -- there are a lot of code changes and as you can see from the commit history I'm not perfect... there have been bugs that I found and corrected.  So, I am keen for other eyes to look this over.

Thanks your interest and support.

dak.

dkerr64:
by-the-way.... in the above post I stated that I have one server that obtains IPv6 addresses for other servers on my internal network and updates the address to freedns.   How do I do that?  With the following ugly cmdv6= in ddclient.conf...
```
usev6=cmd
cmdv6="{ host=qnap; ping $host -c 1 >/dev/null; ipv6=$(ip addr show dev ens3 | grep \"scope.global\" | sed -e's/^.*inet6 \\([^ ]*\\)\\/.*$/\\1/;t;d'); mac=$(arp -a $host | sed -e's/^.* at \\([^ ]*\\) .*$/\\1/;t;d'); ipv6calc -q --in prefix+mac --action prefixmac2ipv6 --out ipv6addr  $ipv6 $mac 2>/dev/null; }"
```
In this case the internal server has a hostname of "qnap"
If you want to try this on a command line then be sure to remove the backslash's before the quotes and backlashes.  In other words...

`{ host=qnap; ping $host -c 1 >/dev/null; ipv6=$(ip addr show dev ens3 | grep "scope.global" | sed -e's/^.*inet6 \([^ ]*\)\/.*$/\1/;t;d'); mac=$(arp -a $host | sed -e's/^.* at \([^ ]*\) .*$/\1/;t;d'); ipv6calc -q --in prefix+mac --action prefixmac2ipv6 --out ipv6addr  $ipv6 $mac 2>/dev/null; }`

Try that from your linux cmd line (substitute valid local host name for qnap) !

dak

wimpunk:
Sorry I missed this one... I'll try to take a review asap.
Did anyone tried the patch already?

wimpunk:
Seems like there are issues when trying to merge and I have some reviews which didn't got any respons.  Can you review them?

dkerr64:
Working on re-syncing with your current master, don't merge until I complete and test please.  I'll let you know once that is done.
Please point me at the review/comments that I have not responded to.

wimpunk:
About the comments: there was a `load_ssl_support` removed from line 1981 and the identification changed.  I argued that I'm aware of a bad indentation but there is no need to correct it on rows you didn't change the contents on.

dkerr64:
I have fixed all the white space diffs.

ddclient
```diff
@@ -1991,74 +2091,67 @@ sub geturl {
     if ( $force_ssl || ($globals{'ssl'} and (caller(1))[3] ne 'main::get_ip') ) {
         $use_ssl      = 1;
         $default_port = 443;
-		load_ssl_support;
```
dkerr64: This `load_ssl_support` is deleted here and moved down to where it is needed.  IPv6 only works with cURL so this should be loaded only if required for IPv4 and only if cURL is not being used for HTTP.

ddclient
```diff
@@ -1991,74 +2091,67 @@ sub geturl {
...
+        ## make sure newlines are <cr><lf> for some pedantic proxy servers
+        ($rq = $request) =~ s/\n/\r\n/g;
+
+        # local $^W = 0;
+        $0 = sprintf("%s - connecting to %s port %s", $program, $peer, $port);
+        if (! opt('exec')) {
+	       debug("skipped network connection");
+	       verbose("SENDING:", "%s", $request);
+        } elsif ($use_ssl) {
+		load_ssl_support;
```
dkerr64: `load_ssl_suppor` moved here... where required.

dkerr64:
`load_ssl_support` was moved lower into the code.  See inline comments.

dkerr64:
An additional comment on `load_ssl_support`.  We ran into several problems using the perl IO::Socket with IPv6 and SSL eventually I had to abandon using it and so if you use IPv6 in ddclient (this pull request) then you need to use cURL... which has to be enabled in the .conf file -- which enables it for both IPv6 and IPv4.  After further review we decided to abandon IO:Socket completely and for the astinux-project pulled all the code out of ddclient in our own fork (https://github.com/astlinux-project/ddclient-curl).   cURL is much more robust and portable across systems and has proven much more stable.  With this pull request I have had to weave in IPv6 support using cURL without breaking any of your existing IPv4 code that uses IO:Socket making the end result more complicated than it need be.

ddclient
```diff
@@ -4149,48 +4415,87 @@ sub nic_nsupdate_update {
...
+					success("updating %s: %s: IPv6 address set to %s", $_, $status, $ipv6);
+				}
+			} else {
+				foreach (@hosts) {
+					failed("updating %s", $_);
+				}
+			}
+
 		}
 	}
```
erul92: load_ssl_support

nirgal:
Hi!

This is great work. I really need ipv4+ipv6 support for http://njalladnspotetti.onion dyndns service: It requires setting up ipv4 and ipv6 at the same time; Setting only one will remove the other.

Did you try to create a push request upstream? (source forge)

dkerr64:
@nirgal see here... https://github.com/astlinux-project/ddclient-curl
Unfortunately @abelbeck and I had to create our own fork for our project as @wimpunk did not act on our PR and we needed this and other changes.

nirgal:
I'm reviewing the PR, in case wimpunk wakes up. So far my only interrogation is regarding the non-default curl option "--insecure". I'm not sure I want that in my ddclient. Can you comment on this? I suppose this disables certificate checks and this is required for some routers? How is that different from situation before your commit?

dkerr64:
@nirgal you are correct, this flag stops CURL from checking that the certificate is valid.  I don't recall exactly why I added this but I assume it is because I ran into cases where we are communicating with sites that have self signed certificates.  It would not be a bad idea to make this an option so that it is only turned on if the user runs into an issue and needs it.

Duffyx:
Hello, has this been merged into the repository? As the PR is still open. I would very much like decent IPv6 support on a dynamic DNS client. Thanks!

dkerr64:
We created our own fork because ddclient owner never merged this.  See here...
https://github.com/astlinux-project/ddclient-curl

Thanks
DAK

dominikholler:
Until ddclient supports dual-stack, I use ddclient's postscript hook to update the IPv6 by curl manually, see https://gist.github.com/dominikholler/aec112a2ee05288588895120f41189d7
Maybe this workaround helps other people, too.

ffries:
It is sad to see that IPv6 is not ddclient priority. 
Is the project dead?

vfreex:
+1 for this feature.
Most Chinese ISPs are supporting IPv6 now but unfortunately we don't have enough ddns support yet.

SuperSandro2000:
I solved the merge conflicts with this branch here https://github.com/SuperSandro2000/ddclient .

Please test it as it could be buggy.

Also @wimpunk @DaveSophoServices if you don't have enough time to maintain this project I would love to help you and get all this stuff reviewed and merged.

wimpunk:
@SuperSandro2000 
I just saw you want to help us supporting ddclient. Are you already subscribed to the mailing list at https://sourceforge.net/p/ddclient/mailman/? Let's discuss what you can do on that list.

dachaac:
FYI:
Just received a newsletter from no-ip.com and they are now finally providing IPv6 support for their service. They claim that they are pretty close to "industry standard". But I suppose you know that better.

Ref:
https://www.noip.com/integrate
https://www.noip.com/integrate/request
https://www.noip.com/integrate/add-noip

Doesn't look too complicated once base IPv6 support is in place. Is there some help needed to get this in or just waiting it to happen?

dkerr64:
There are too many merge conflicts with the current version of ddclient.  I will bring the IPv6 support up-to-date with current master and then resubmit a new pull request.  You can follow along here if you are interested.... https://github.com/dkerr64/ddclient/tree/IPv6
Thanks.

PR #41, opened 2016-05-13
Use ipify as default web service

michiil:
Since many users including myself had problems with http://checkip.dyndns.org/ i suggest moving to ipify as default service
Info: https://www.ipify.org/
Pro:
-No limit in requests
-Very fast and responsive
-Open source
-https
-No Logging of visitors
-Works with IPv4 and IPv6

Con:
-?

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -304,6 +304,7 @@ sub ip_strategies_usage {
 }
 
 my %web_strategies = (
+	'ipify'=> 1,
 	'dyndns'=> 1,
 	'dnspark'=> 1,
 	'loopia'=> 1,
@@ -333,7 +334,7 @@ my %variables = (
 	'ip'                  => setv(T_IP,    0, 0, 1, undef,                undef),
 	'if'                  => setv(T_IF,    0, 0, 1, 'ppp0',               undef),
 	'if-skip'             => setv(T_STRING,1, 0, 1, '',                   undef),
-	'web'                 => setv(T_STRING,0, 0, 1, 'dyndns',             undef),
+	'web'                 => setv(T_STRING,0, 0, 1, 'ipify',             undef),
 	'web-skip'            => setv(T_STRING,1, 0, 1, '',                   undef),
 	'fw'                  => setv(T_ANY,   0, 0, 1, '', 		      undef),
 	'fw-skip'             => setv(T_STRING,1, 0, 1, '',                   undef),
@@ -373,7 +374,7 @@ my %variables = (
 	'use'                 => setv(T_USE,   0, 0, 1, 'ip',                 undef),
 	'if'                  => setv(T_IF,    0, 0, 1, 'ppp0',               undef),
 	'if-skip'             => setv(T_STRING,0, 0, 1, '',                   undef),
-	'web'                 => setv(T_STRING,0, 0, 1, 'dyndns',             undef),
+	'web'                 => setv(T_STRING,0, 0, 1, 'ipify',             undef),
 	'web-skip'            => setv(T_STRING,0, 0, 1, '',                   undef),
 	'fw'                  => setv(T_ANY,   0, 0, 1, '', 		      undef),
 	'fw-skip'             => setv(T_STRING,0, 0, 1, '',                   undef),
--- a/ddclient
+++ b/ddclient
@@ -66,6 +66,7 @@ sub T_POSTS	{'postscript'};
 
 ## strategies for obtaining an ip address.
 my %builtinweb = (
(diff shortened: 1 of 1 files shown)


wimpunk:
Looks like a good suggestion.  I'll review your patch asap.

dkerr64:
I tested this to see if it works with ipv6 and it fails...

```
~ # curl --ipv4 --url 'http://api.ipify.org'
50.xx.yy.zz 
~ # curl --ipv6 --url 'http://api.ipify.org'
curl: (6) Couldn't resolve host 'api.ipify.org'
```

It is fine to add it to ddclient, but it is not true to assert that it works with ipv6.
Very few IP discovery websites actually support ipv6.  One that does is...

```
~ # curl --ipv6 --url 'https://ipv6.nsupdate.info/myip'
[returns my IPv6 address]
```

David

michiil:
Hmm okay thanks for trying this.. I couldn't try it since i dont have an IPv6 address..
So you say myip.nsupdate.info would be fine? I could open a new pr..
But getting away from dyndns as default is important imo.

Michael

geraldhansen:
I can recommend myonlineportal.net.
This will work with ipv6 and ipv4 and they support https too.
`curl --ipv6 https://myonlineportal.net/checkip`
`curl --ipv4 https://myonlineportal.net/checkip`

Gerald

michiil:
Well i can't decide whats best to use..
I will just make a nwe pr as soon as we find a fitting service..

Mjcaehl

dkerr64:
Gerald, thanks I did not know of myonlineportal.net.  This is the first site I've found that seems to work well for both IPv6 and IPv4.

Note that ddclient as published does not handle IPv6, what limited support that was added is very specific and doesn't do the job.  My pull request #40 does add support for IPv6 in a manner that I think is flexible and comprehensive.  But I realise that the changes are considerable -- I've requested a code review, testing and comment. 

The concept of Dynamic DNS / IP doesn't really have a place in the IPv6 world.  And few DDNS sites support it.  FreeDNS does, but I think there are problems with it too.  However I do like the idea of a DDNS client that can update IPv6 addresses, even though the actual address is highly unlikely to ever change unless the server/client system itself changes.  But that DDNS client needs to be able to cope with both IPv4 and IPv6 addresses for the same domain name, and that requires a lot of changes to ddclient, hence the size of pull request #40 

David

michiil:
So if we don't support IPv6 ipify would be fine untill we do..
I still think it would be the best option because they explicitly memtion no limit on requests and thats pretty cool imo..

Michael

wimpunk:
The problem with IPv6 has already been reported.  Let's all follow the bug and hope there will be a good ipv6 provider soon.  See https://github.com/rdegges/ipify-api/issues/3

dkerr64:
I would observe that in IPv6 world there is really no need for web discovery of your IP address.  Even behind NAT the IPv6 address assigned to your network interface is globally routable so you can obtain your IPv6 address simply by querying the network interface (eth0 or whatever).  So for ddclient purposes it is  unnecessary.

Web discovery is useful if, for example, you want to find out what your browser is using by default.

David

risturiz:
Why not -> http://checkip.dns.he.net/ ?

petersjt014:
They seem to have added IPv6 support. Testing it [here](http://www.ipv6proxy.net/go.php?u=https%3A%2F%2Fapi6.ipify.org%2F%3Fformat%3Djson&b=0&f=norefer) seems to indicate it works, I'll test it on a vm w/ IPv6 when I have time.

xlammertink:
@SuperSandro2000: I see you added the ipify web-service using the URL https://api.ipify.org, but you might as well want to add ipify6 web-service using URL https://api6.ipify.org as @petersjt014 mentioned.

SuperSandro2000:
> @SuperSandro2000: I see you added the ipify web-service using the URL [api.ipify.org](https://api.ipify.org), but you might as well want to add an ipify6 web-service using URL [api6.ipify.org](https://api6.ipify.org) as @petersjt014 mentioned.
> 
> Additionally dnspark is not working anymore as the domain name is for sale. Maybe remove that one as well?

I just added it to my fork cause I needed some ipv6 features and decided to just merge everything good together.

petersjt014:
> @SuperSandro2000: I see you added the ipify web-service using the URL https://api.ipify.org, but you might as well want to add an ipify6 web-service using URL https://api6.ipify.org as @petersjt014 mentioned.
> 
> Additionally dnspark is not working anymore as the domain name is for sale. Maybe remove that one as well?

I just remembered that I listed all the dead ones I could find in #87.

PR #43, opened 2016-07-30
Email protocol

iboisvert:
For people with internet-facing services that don't need or want a dynamic dns account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Dynamic DNS services currently supported include:
     Google      - See http://www.google.com/domains for details
     Duckdns     - See https://duckdns.org/ for details
     woima.fi    - See https://woima.fi/ for details
+    email       - Sends an email notification to specified recipients
 
 DDclient now supports many of cable/dsl broadband routers. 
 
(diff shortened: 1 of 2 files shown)


wimpunk:
I think I understand why you want this kind of feature.  Unfortunately your addition uses to much extra dependencies.  Since we are trying to add as less dependencies as possible, I believe it would be better to add an 'none' provider and configure a mail-option. That would add less to the current code.

dkerr64:
Maybe a better way to do this would be to provide for a way to execute a command.  Right now ddclient lets you execute a command to obtain the IP address.   How about adding a feature to execute a command when an update is performed... passing in the obtained IP address (and maybe other flags like whether IPv6, what service was updated, etc.) as parameters.  In this case that command could be sendmail (wrapped in some shell script).  And it would add tremendous flexibility for other features as well.   As command execution is already supported this could be done with no additional dependencies.

iboisvert:
Hi, thanks for sharing your thoughts. I agree, an email option would be
better than the provider, then it could be used conjunction with any of the
other providers as well. One other thing that I thought of, though, is that
there needs to be an additional option to specify a certificate to sign the
generated email, because these emails would be very easy to spoof.

Best regards,
Ian

wimpunk:
@dkerr64 The option of running a script does already exist.  See the `postscript` option for more info.

PR #44, opened 2016-09-08
Cloudflare Config

esetnik:
Specifying the `server` directive caused the Cloudflare protocol to use the wrong url when communicating with Cloudflare API v4. It works if the directive is removed.

Changes:
- Update sample config with functional Cloudflare settings.


wimpunk:
Thansk!  I always like it when some reads and fixes the documentation.

PR #45, opened 2016-11-04
Example config showing how to retrieve IP from fritzbox via UPnP using ddclients cmd option

Rusk85:
Pull requests consists out of 2 files. 
1) ddclient.conf stripped to bare minimum with required changes to use script output as source of IP
2) get-ip-from-fritzbox as the script that is the source of the IP


wimpunk:
Your request has partial be used in e33aa2fa96d7ef15 and 40409911b33f since your example wasn't really related to DuckDNS.

SuperSandro2000:
I think this can be closed.

PR #47, opened 2017-01-16
Adding support for free dynamic dns freemyip.com

Cadence-GitHub:
I tested to locally and it seems to be working fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Dynamic DNS services currently supported include:
     CloudFlare  - See https://www.cloudflare.com/ for details
     Google      - See http://www.google.com/domains for details
     Duckdns     - See https://duckdns.org/ for details
+    Freemyip    - See https://freemyip.com for details
     woima.fi    - See https://woima.fi/ for details
 
 DDclient now supports many of cable/dsl broadband routers. 
--- a/README.ssl
+++ b/README.ssl
@@ -8,3 +8,4 @@ On alpine, you need perl-io-socket-ssl to have IO::Socket::SSL
 
 ssl support is tested on folowing dynamic dns providers:
 - dyndns.com
+- freemyip.com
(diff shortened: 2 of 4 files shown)


wimpunk:
Your patch looks great, thanks for pushing it.
Just a little question: do I see it correctly there is no login and the IP isn't fetched from anywhere?

Cadence-GitHub:
@wimpunk Thanks for your comment.
Yes, you are correct. There is no login at freemyip.com, and the service detects client's IP address automatically, so there is no need to fetch IP.

I did some testing with updating from different IPs, and it worked correctly. I also tried some negative-case tests (wrong token, no token at all, etc) and the results and errors were as expected.

Cadence-GitHub:
Hi @wimpunk 
Is there anything else I should be doing regarding this PR? I'm quite new to github, so am not very familiar with the whole PR procedure. If you still require anything from me, please let me know :)

wimpunk:
Sorry, didn't had time to review and test it.

wimpunk:
Found some time to review your suggestion.  Merged. :-)

Cadence-GitHub:
That's great news, thanks @wimpunk. So exciting to see my code added to ddclient!

glensc:
seems to be ddclient itself is unable to operate if it can't detect ip by itself. i had to configure `use=web` for ddclient even to attempt to submit registration

glensc:
btw, their api supports passing ip as well:
- http://freemyip.com/help?domain=example.freemyip.com&token=exampletoken

> ### Can I point my domain to an arbitrary IP address?
> Yes. You can use an optional "&myip" parameter to do that. For example, to set your domain to point to IP 1.2.3.4 you need to use the following URL:
> ```
> https://freemyip.com/update?token=exampletoken&domain=example.freemyip.com&myip=1.2.3.4
> ```

PR #49, opened 2017-05-04
Interpret 'Address not changed' reply from freedns and tidy up indents.

hattiecat:
Freedns appears to have changed the response when an update attempt is made but the address has not changed, returning an error. This interprets the error response.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -3841,6 +3841,12 @@ sub nic_freedns_update {
 		$config{$h}{'mtime'}  = $now; 
 		$config{$h}{'status'} = 'good'; 
 		success("updating %s: good: IP address set to %s", $h, $ip); 
+	    } elsif ($reply =~ /ERROR: Address (.*) has not changed\./) {
+		$ip = $1;
+		$config{$h}{'ip'}     = $ip; 
+		$config{$h}{'mtime'}  = $now; 
+		$config{$h}{'status'} = 'good'; 
+		success("update not necessary %s: IP address already set to %s", $h, $ip); 
 	    } else {
 	        $config{$h}{'status'} = 'failed';
 		warning("SENT: %s", $freedns_hosts{$h}->[2]) unless opt('verbose');


dkerr64:
How is this ever getting executed?  There is a test a few lines above to check whether the IP address at FreeDNS is already set to the requested IP address if so then it skips any attempt to update it.  So in theory there should never be a case where ddclient goes and attempts to update the IP address when it is already set to the right value.  Do you have a test case where ddclient was failing?

dkerr64:
While examining this code I have identified another possible issue.  FreeDNS expires URLs/IPs that have not been updated in 30 days.  There is no logic in ddclient to take care of this.  So what would appear to be happening is that ddclient happily checks to see if IP address is set and if so skips the update.  Eventually FreeDNS will expire the URL/IP and the next time ddclient tests it will find that there is no entry for the URL and will then proceed to set it.   So there will be a gap, albeit a small one, when the IP address is not set for a URL.

hattiecat:
I was using ddclient version 3.8.2 on Raspbian (Debian 8 derived) with FreeDNS and the freedns protocol. When ddclient started it successfully it hit the server for the IP address and got it, then immediately tried to fetch it again and failed, which cleared the ip address. syslog below. This worked 6 months ago so I can only guess that FreeDNS have done some change to the protocol.

May  7 06:46:01 foobar ddclient[923]: WARNING:  file /var/cache/ddclient/ddclient.cache, line 3: Invalid Value for keyword 'ip' = ''
May  7 06:46:02 foobar ddclient[923]: WARNING:  SENT: http://freedns.afraid.org/dynamic/update.php?**********
May  7 06:46:03 foobar ddclient[923]: WARNING:  REPLIED: HTTP/1.1 200 OK
May  7 06:46:03 foobar ddclient[923]: WARNING:   Server: nginx
May  7 06:46:03 foobar ddclient[923]: WARNING:   Date: Sat, 06 May 2017 21:46:02 GMT
May  7 06:46:03 foobar ddclient[923]: WARNING:   Content-Type: text/plain; charset=utf-8
May  7 06:46:03 foobar ddclient[923]: WARNING:   Connection: close
May  7 06:46:03 foobar ddclient[923]: WARNING:   Vary: Accept-Encoding
May  7 06:46:03 foobar ddclient[923]: WARNING:   Cache-Control: no-store, no-cache, must-revalidate
May  7 06:46:03 foobar ddclient[923]: WARNING:   Cache-Control: post-check=0, pre-check=0
May  7 06:46:03 foobar ddclient[923]: WARNING:   Pragma: no-cache
May  7 06:46:03 foobar ddclient[923]: WARNING:   Expires: Mon, 26 Jul 1997 05:00:00 GMT
May  7 06:46:03 foobar ddclient[923]: WARNING:   X-Cache: EXPIRED
May  7 06:46:03 foobar ddclient[923]: WARNING:
May  7 06:46:03 foobar ddclient[923]: WARNING:   ERROR: Address ***.***.***.*** has not changed.
May  7 06:46:03 foobar ddclient[923]: FAILED:   updating ***.***.***: Invalid reply.

hattiecat:
I can try backing out the change, restarting and send you the log, but it'll be later.

dkerr64:
Could you try the version of ddclient proposed in PR #40 it uses the newer "v2" API protocol for FreeDNS.  I'm curious if you have the same problem with it.
Thanks
David

dkerr64:
And/or try the version of ddclient at https://github.com/astlinux-project/ddclient-curl which has had a number of cleanup/enhancements made... most notably to exclusively use cURL for all http requests, dumping the perl network APIs which have a number of issues when trying to talk IPv6. I'd be curious if that version has the same issue.
Thanks
David

hattiecat:
> How is this ever getting executed? There is a test a few lines above to check whether the IP address at 
> FreeDNS is already set to the requested IP address if so then it skips any attempt to update it. 

Here's a log file. I've redacted the IP address and a few other details. In the response to the first request to the server  
GET /api/?action=getdyndns&sha=****** HTTP/1.0 
at line 20, it says:
INFO:     setting IP address to 192.168.10.109 for foo.bar.baz

This is the LOCAL address of that interface NOT the address returned in the response. It then fires off an update request
 GET /dynamic/update.php?aWE5NWp0cXdrOWV0ZG9oeWEzaEd2VmQ4OjE2MTg1OTU2 HTTP/1.0

to which the server responds
ERROR: Address ***.***.***.*** has not changed.

which can't be grokked.


[log.txt](https://github.com/wimpunk/ddclient/files/983192/log.txt)

hattiecat:
Perhaps I should add that the machine is sitting behind my ISP's router which is using NAT. It's on the "DMZ" IP address, so external incoming TCP connections get sent to its address (192.168.10.109). It's puzzling that the first response from the FreeDNS server contains the public IP address but the INFO line says "setting IP address to 192.168.10.109 for foo.bar.baz" - where is it getting the local interface's IP address from?

hattiecat:
> Could you try the version of ddclient proposed in PR #40 it uses the newer "v2" API protocol for 
> FreeDNS. I'm curious if you have the same problem with it.

Hi David. I'd be happy to but am not /au fait/ with Github (this is my first "pull request", or indeed, my first repository!) so I'm not sure how to download it. If you would be so kind as to send me the source by email (hattiecat@lowsnr.net) as a compressed tar archive or something I'll give it a whirl.

hattiecat:
When restarting with the modded version I sometimes get an error:
$ sudo rm /var/cache/ddclient/ddclient.cache 
$ sudo /usr/local/bin/ddclient --verbose
INFO:     forcing updating foo.bar.baz because no cached entry exists.
CONNECT:  freedns.afraid.org
WARNING:  cannot connect to freedns.afraid.org:80 socket: IO::Socket::INET: Bad hostname 'freedns.afraid.org'
FAILED:   updating foo.bar.baz: Could not connect to http://freedns.afraid.org/api/?action=getdyndns&sha=**** for site list.

Any idea why?

dkerr64:
You are trying to set an invalid IP address... I would expect FreeDNS to reject any IP in the local address range defined by [RFC 1918](https://tools.ietf.org/html/rfc1918). You need to establish your external IP address yourself or change your ddclient.conf file to use a different method to establish your external IP address.  It looks like you are using
`use=if`
which won't work for a system behind NAT and unless the system is externally connected you should
```
use=web
web=<one of the external sites that returns your IP address, or let ddclient pick>

```
Arguably the flaw in ddclient is not detecting this before trying to update the remote DNS service, whether this is deliberate or not I don't know but a warning message at a minimum might be a good thing to do (I suppose there could be a use case where you do in fact want to update a dynamic DNS server that is also behind the same NAT as the client system).

hattiecat:
Ah, that's fixed it! (Embarrassed.) Thanks.

In my defence I'd argue that the documentation is not very explicit and there are no examples given regarding 'use=web' even though this may be a very common use case - i.e. people use dynamic DNS because they have a home system behind a NAT provided by an ISP.

I probably won't be the last person to make this mistake so adding a warning and/or improving the 'robustness' (e.g. my submitted patch should work if only for the wrong reason!) might save some pain for someone else.

dkerr64:
@hattiecat thank you for confirming.   Your fix is specific to FreeDNS, it would likely be a lot of work to extend to all providers.  I think a warning from ddclient to alert user to possible configuration error may be best as that could be done in the common code.

hattiecat:
Sorry for causing any bother, and thanks for your help.

I clearly don't understand the use cases for Dynamic DNS very well, but perhaps the common ones are:

(1) Obtain global IP address locally (e.g. from the router), and notify the Dynamic DNS server of the address to allow it to update the DNS entry. In this case, set one of the following
-use=if, if interface
-use=ip,-ip address
-use=fw, fw address|url
-use=<router>

(2) Obtain the IP address by sending an http request to the Dynamic DNS server itself, which then automatically updates the DNS entry in one step. In this case, set
-use=web[, -web provider|url] 

If the above is correct, then working something similar into the '== CONFIGURING ddclient' part of the help and in the README.md file might make things a little clearer.

dkerr64:
I will leave it up to @wimpunk to decide what action to take whether in documentation or other.  You ran across a situation where an error condition was probably the correct response as it indicated a configuration error that once resolved, ddclient worked properly for you.

wimpunk:
@dkerr64 Any suggestion in changing the readme will be considered, especially if formed as a pull request.

wimpunk:
@hattiecat Your suggestion looks good. I'll first try it out before officially merging.

dkerr64:
@wimpunk I wouldn't change anything.  This was a case of user attempting to set a invalid IP address at a dynamic DNS provider.  Most people using ddclient would be aware that this is not valid.

ddclient
```diff
@@ -3841,6 +3841,12 @@ sub nic_freedns_update {
...
 		success("updating %s: good: IP address set to %s", $h, $ip); 
+	    } elsif ($reply =~ /ERROR: Address (.*) has not changed\./) {
+		$ip = $1;
+		$config{$h}{'ip'}     = $ip; 
+		$config{$h}{'mtime'}  = $now; 
+		$config{$h}{'status'} = 'good'; 
+		success("update not necessary %s: IP address already set to %s", $h, $ip); 
 	    } else {
 	        $config{$h}{'status'} = 'failed';
 		warning("SENT: %s", $freedns_hosts{$h}->[2]) unless opt('verbose');
```
erul92: Unless

erul92: 2 unless

SuperSandro2000:
I am going to close this cause it was probably a configuration mistake.

PR #50, opened 2017-05-13
Support obtaining interface name from default route

jikamens:
Allow the user to specify "ddclient" to the "if" option to tell
ddclient to run `route -n` and look at the default route in its output
to determine which interface to query for the IP address.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -710,6 +710,7 @@ my @opt = (
     [ "ip",          "=s", "-ip address           : set the IP address to 'address'" ],
     "",			     
     [ "if",          "=s", "-if interface         : obtain IP address from 'interface'" ],
+                         "                          Specify 'defaultroute' to obtain interface name from default route.",
     [ "if-skip",     "=s", "-if-skip pattern      : skip any IP addresses before 'pattern' in the output of ifconfig {if}" ],
     "",
     [ "web",         "=s", "-web provider|url     : obtain IP address from provider's IP checking page" ],
@@ -2164,10 +2165,22 @@ sub get_ip {
 	$arg = 'ip';
 
     } elsif ($use eq 'if') {
+        my $iface;
 	$skip  = opt('if-skip', $h)  || '';
-	$reply = `ifconfig $arg 2> /dev/null`;
-	$reply = `ip addr list dev $arg 2> /dev/null` if $?;
-	$reply = '' if $?;
+        if ($arg eq 'defaultroute') {
+            ($iface) = (grep(/^0\.0\.0\.0\b/, `route -n`))[0] =~ /(\S+)$/;
+        }
+        else {
+            $iface = $arg;
+        }
+        if (! $iface) {
+            $reply = '';
+        }
+        else {
+            $reply = `ifconfig $iface 2> /dev/null`;
+            $reply = `ip addr list dev $iface 2> /dev/null` if $?;
+            $reply = '' if $?;
+        }
 
     } elsif ($use eq 'cmd') {
 	if ($arg) {


jikamens:
This is useful for people who switch among different interfaces, e.g., a laptop which is sometimes connected to the internet via wifi and sometimes via a cable.

dkerr64:
Could the same not be accomplished using the cmd= option?  Certainly not as straight forward as adding support into ddclient, but certainly doable.

jikamens:
>Could the same not be accomplished using the cmd= option?

Yes.

>Certainly not as straight forward as adding support into ddclient, but certainly doable.

I submitted this PR exactly because it's "not as straightforward as adding support into ddclient." It seems to me that this is a sufficiently common workflow that it is worth having ddclient support it directly rather than requiring people to figure out how to implement it themselves with shell commands or scripts.

wimpunk:
A few comments:
* I think most people use the `web` configuration for situations like this.  
* I don't understand why your laptop should need a public DNS name, could you explain the use of that
* route will *not* work for all operating systems.  A few distributions will need the complete path while others are using the `ip route` command to find the routing information.
So if you can provide a solution for the other situations and explain me why it is useful, I'll add it to ddclient.

jikamens:
These are very reasonable questions. I have (what I hope are) very reasonable answers. :-)

>I think most people use the `web` configuration for situations like this.

The `web` configuration won't work because I am trying to register in DNS the _internal_ IP address of my laptop, not the public IP address. I run `sshd` on my laptop and I want to be able to connect to it from other machines on the LAN as needed, but since it uses DHCP and I connect it to multiple LANs (home, work), its IP address changes on a regular basis.

(Most common user story for me: I am sitting in my basement office working on my desktop computer when I realize that there's a file I need on my laptop, which is plugged in two floors up in my bedroom. Rather than walking up two flights of stairs to get the file, I use `scp` to retrieve it directly.)

>I don't understand why your laptop should need a public DNS name, could you explain the use of that

See above.

>route will _not_ work for all operating systems. A few distributions will need the complete path while others are using the `ip route` command to find the routing information.
>So if you can provide a solution for the other situations and explain me why it is useful, I'll add it to ddclient.

Yes, true. But I don't have access to all those other operating systems so I can't implement and test solutions for them. I was sort of hoping that, should this functionality prove useful to others as it is for me, the people using those other operating systems who want to use this functionality could submit PRs to improve it for greater compatibility.

("Don't let the perfect be the enemy of the good.")

dkerr64:
@jikamens you are trying to set an _internal_ IP address at an _external public_ Dynamic DNS site ?  That is likely not going to work very well... if you are at your office you will not be able to access your bedroom across the internet using an internal IP address as destination.  And many Dynamic DNS site will reject attempt to set a local (non-routable) IP address (there is a parallel discussion around this in PR #49 
If the problem is accessing your laptop within your home then use the hostname of the laptop, not its IP address.  Your local network will route that to whatever IP address your DHCP server assigned. All devices have a hostname.... on a Mac you can find it in system settings under "sharing"

DAK

jikamens:
Dude.

I've been doing system administration for 30 years. I know what I'm doing.

The way I'm doing things now works just fine. None of the DDNS providers I'm using have any objection to my registering a 192.168.x.x IP address in my DNS records.

We don't use any sort of mDNS infrastructure either at home or at work, nor does the DHCP server at home or at work do any sort of DDNS registration in any sort of local DNS domain. So no, the host name that my laptop reports to DHCP is not in fact accessible anywhere other than where I put it with ddclient.

The functionality I've submitted here solves a real problem, is useful for me, and works for me. I suspect that it would be useful and work for other people as well. I think it is unlikely that I'm the only person in the entire world who wants to do what I've shown here. But whatever, I'm done wasting my time trying to sell it to you. Take it or close the PR, your choice.

ddclient
```diff
@@ -2164,10 +2165,22 @@ sub get_ip {
 	$arg = 'ip';
 
     } elsif ($use eq 'if') {
+        my $iface;
 	$skip  = opt('if-skip', $h)  || '';
-	$reply = `ifconfig $arg 2> /dev/null`;
-	$reply = `ip addr list dev $arg 2> /dev/null` if $?;
-	$reply = '' if $?;
+        if ($arg eq 'defaultroute') {
+            ($iface) = (grep(/^0\.0\.0\.0\b/, `route -n`))[0] =~ /(\S+)$/;
```
andhe, line 2171: A fallback to use `ip route ...` here would be useful (like all the other places a net-tools command is executed).

eg.
`ip -j -p route show default | grep '"dev":' | cut -d\" -f4`

dkerr64: There are at least four ways to get the default route / interface that need to be handled.  I have an implementation in a fork that I have been working on.  I plan to submit it (not the entire fork, just parts of it) as a PR on ddclient once some of my other pre-req PRs are merged. https://github.com/dkerr64/ddclient/blob/a99d3333cb70bb6c77a97948ad57ca112eb0e1d8/ddclient.in#L2518

Using -j is an interesting thought, but unfortunately not available on all platforms / versions of iproute2.

SuperSandro2000:
Closing in favor of #266.

PR #53, opened 2018-03-07
Fixing SSL_verify_mode deprecated warning.

digrouz:
Fixing SSL_verify_mode deprecated warning:
*******************************************************************
 Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
 is deprecated! Please set SSL_verify_mode to SSL_VERIFY_PEER
 together with SSL_ca_file|SSL_ca_path for verification.
 If you really don't want to verify the certificate and keep the
 connection open to Man-In-The-Middle attacks please set
 SSL_verify_mode explicitly to SSL_VERIFY_NONE in your application.
*******************************************************************

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,97 @@
+2018-08-09  wimpunk
+
+        * [r208] ddclient: cosmetic, remove stray space indent
+        * [r207] ddclient: Support IPv6 for CloudFlare
+        * [r206] ddclient: name cheap support https now
+
+          From name cheap it seems http is supported now.
+          Since the password was send on plaintext, https should be used
+
+        * [r205] ddclient: Use JSON::PP instead of the (deprecated)
+          JSON::Any
+        * [r204] ddclient: Follow expected behavior
+
+          Align ddclient behavior and documentation with namecheap's -
+          https://www.namecheap.com/support/knowledgebase/article.aspx/583/11/how-do-i-configure-ddclient
+
+        * [r203] ddclient: Specify port number properly to 'nsupdate' (#58)
+
+          If a port number is included in the 'server' configuration item,
+          ddclient allows a port number to be specified by appending a
+          colon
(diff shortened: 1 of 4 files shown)


wimpunk:
The idea is good but fortunately there are mis-configured SSL servers out there.  Nowadays most providers are configured correctly but routers mostly have wrong certificates. The user should still have the possibility to use the SSL_VERIFY_NONE when he knows what he's doing.  He should also have the possibility to specify the location of the ca_file or ca_path.

digrouz:
You mean I need to add options for SSL_VERIFY_NONE and for ca_path ?

DaveSophoServices:
Will try to implement this with a flag for disabling the SSL_VERIFY code.

SuperSandro2000:
This is not shown for me on Ubuntu. Did the package change this behavior and deprecated the default mode?

mike2307:
I'm using RHEL 7 with the package perl-IO-Socket-SSL.noarch 1.94-7.el7

I get the same warning. To fix it, it have to add
`SSL_verify_mode => 1,`
instead.

It won't let me use a string or a constant.

dkerr64:
I believe that this would be fixed with PR #233 which includes new option for `-{no}web-ssl-validate` (in both the curl and non-curl code path).

SuperSandro2000:
Closing in favor of #233

dkerr64:
Shouldn't the issue remain open until after the PR is merged?

SuperSandro2000:
Well, maybe but it is a pull request which isn't mergable since a long time and the opener did not respond, so I think we can keep it closed.

dkerr64:
> Well, maybe but it is a pull request which isn't mergable since a long time and the opener did not respond, so I think we can keep it closed.

Ah... my mistake, I thought this was an issue but is in fact a PR.  So yes, close out old non-mergeable PRs in favor of the one that can merge.

PR #57, opened 2018-06-02
 Fix sample Cloudflare config

Denton-L:
The extra comma bleeds into the API request sent to Cloudflare and, as a
result, causes it to fail. Removing the comma fixes this problem.


wimpunk:
Actually the comma is needed if you also specify the TTL.  Maybe we should change the order and end with the password line since that line is always needed

Denton-L:
Thanks for your feedback. I've updated the commit to reflect this.

PR #58, opened 2018-07-21
Specify port number properly to 'nsupdate'

kpfleming:
If a port number is included in the 'server' configuration item,
ddclient allows a port number to be specified by appending a colon
and the port number to the server's name or IPv4 address. However,
nsupdate does not support this syntax, it requires the port number
to be separated from the server name/address by whitespace.

Signed-off-by: Kevin P. Fleming <kevin@km6g.us>


wimpunk:
Thanks for the fix!

PR #62, opened 2019-01-05
Update Cloudflare zone ID query

mkw:
The zone ID query was executed without a trailing slash, resulting in an un-handled 301.


mkw:
Never mind -- misconfigured the base URL; oddly, this did make it get further, but it did not completely fix it.

PR #63, opened 2019-01-21
Patch 1

matthewpayne:
Included support for dnsmadeeasy.com
it's the same as sitelutions only changes are: https://cp.dnsmadeeasy.com/servlet/updateip? instead


wimpunk:
Patch looks good.  I'll try to add it asap.

wimpunk:
What's the difference with #60?

okeeblow:
#60 supports TLS and has full error code->message handling where this patch does not.

PR #65, opened 2019-03-11
Add perl modules to install instructions

SuperSandro2000:


wimpunk:
json is only needed with cloudfare and freedns.  If you could add a note
about that I'll add your patch to ddclient.

SuperSandro2000:
I didn't know that as I only ever used cloudflare.

SuperSandro2000:
Just a headsup that I changed the PR.

PR #67, opened 2019-04-06
Fix fritz box example for newer routers

SuperSandro2000:


SuperSandro2000:
Is there any reason why nothing is happening with this?

wimpunk:
Yes, there is a reason.  I currently try to maintain the project but I don't have any time to add new things to the code.

PR #70, opened 2019-08-03
add nsupdate.info config samples

fabianskibr:
add nsupdate.info config samples

### DIFF
--- a/sample-etc_ddclient.conf
+++ b/sample-etc_ddclient.conf
@@ -240,3 +240,24 @@ ssl=yes					# use ssl-support.  Works with
 # login=your-myonlineportal-username
 # password=your-myonlineportal-password
 # domain.myonlineportal.net
+
+## 
+## nsupdate.info IPV4(https://www.nsupdate.info)
+##
+#protocol=dyndns2
+#use=web, web=http://ipv4.nsupdate.info/myip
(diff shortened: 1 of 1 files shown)


sample-etc_ddclient.conf
```diff
@@ -240,3 +240,24 @@ ssl=yes					# use ssl-support.  Works with
...
 # password=your-myonlineportal-password
 # domain.myonlineportal.net
+
+## 
+## nsupdate.info IPV4(https://www.nsupdate.info)
+##
+#protocol=dyndns2
+#use=web, web=http://ipv4.nsupdate.info/myip
+#server=ipv4.nsupdate.info
+#login=domain.nsupdate.info
```
DaveSophoServices: how about [yourdomain].nsupdate.info ?

sample-etc_ddclient.conf
```diff
@@ -240,3 +240,24 @@ ssl=yes					# use ssl-support.  Works with
...
+#password='123'
+#domain.nsupdate.info
```
DaveSophoServices: ...likewise [yourdomain].nsupdate.info

fabianskibr: true, it would be better

PR #98, opened 2020-02-05
Fixed cloudflare cache not updating properly

samwell61:
Found that cloudlare wasn't updating the local cache properly after updating the IPs for all my subdomains, it would write the cache values to the first domain in the list and nothing for the rest.


benaud:
@samwell61 Nice, thank you, this fixed a long standing issue for me, appreciated!

PR #117, opened 2020-05-25
Catch SIGINT instead of SIGKILL

rhansen:
SIGKILL can't be caught so there's no point in trying. SIGINT is raised when the user hits Ctrl-C.


rhansen:
Same problem here as PR #116—the commits got squashed together which destroyed the commit messages.

PR #119, opened 2020-05-25
Do not prevent daemonization when force is set

rhansen:
The force option should only be about ignoring the cache and existing value, not about controlling when updates are triggered or whether the program runs in the foreground or not.


rhansen:
Rebased to current master.

SuperSandro2000:
I should merge this probably first or we rebase it a third time. 😂

PR #124, opened 2020-05-27
Add myonlineportal as ip getter

SuperSandro2000:
As suggested here https://github.com/ddclient/ddclient/pull/41#issuecomment-219269685


ddclient
```diff
@@ -303,6 +304,7 @@ sub ip_strategies_usage {
 my %web_strategies = (
```
rhansen, line 304: As far as I can tell `%web_strategies` is not used anywhere. Maybe delete it?

SuperSandro2000: You are right. Ddclient is just the definition of legacy code :)

PR #133, opened 2020-05-27
Only write the PID file if daemonized

rhansen:
There's not much point in writing a PID file if the user does not want to fork. (The process executing ddclient already knows the PID, so it can save it to a file if desired.)

This change is a prerequisite to using [Proc::Daemon](https://metacpan.org/pod/Proc::Daemon) to fix the buggy daemonization logic.


rhansen:
Marked as work in progress because PR #134 should be merged first. I'll rebase this once PR #134 is merged.

PR #135, opened 2020-05-28
Fix usage()

rhansen:
See individual commits for details.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -807,11 +807,7 @@ $result = 'OK';
 
 test_geturl(opt('geturl')) if opt('geturl');
 
-## process help option
-if (opt('help')) {
-    *STDERR = *STDOUT;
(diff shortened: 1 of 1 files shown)


ddclient
```diff
@@ -1407,24 +1406,6 @@ sub init_config {
...
-    $exitcode = shift if @_ != 0;  # use first arg if given
-    my $msg = '';
-    if (@_) {
-        my $format = shift;
-        $msg .= sprintf $format, @_;
-        1 while chomp($msg);
-        $msg .= "\n";
-    }
-    printf STDERR "%s%s\n", $msg, $opt_usage;
-    sendmail();
```
SuperSandro2000, line 1424: Just a note: fatal also does sendmail so we do not lose functionality.

PR #136, opened 2020-05-28
Init script cleanups

rhansen:
This is preparation for some daemonization fixes.

### DIFF
--- a/sample-etc_rc.d_init.d_ddclient
+++ b/sample-etc_rc.d_init.d_ddclient
@@ -33,59 +33,56 @@ export PATH
 
 # See how we were called.
 case "$1" in
-  start)
-	# Start daemon.
-    DELAY=`grep -v '^\s*#' $CONF | grep -i -m 1 "daemon" | awk -F '=' '{print $2}'`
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Tested and works now as expected. Thank you again!

Btw What do you think about organizing all those sample files into some folder structure? I don't know if there is any standard for this that maybe is inline with debian policy or some common perl way to do this.

rhansen:
Thanks for merging!

> What do you think about organizing all those sample files into some folder structure?

I think they're fine where they are, but I wouldn't object to moving them to a `samples/` directory or something like that.

> I don't know if there is any standard for this that maybe is inline with debian policy or some common perl way to do this.

I don't know about Perl conventions. For Debian it doesn't matter where they are; they're just [listed in an `examples` control file](https://salsa.debian.org/debian/ddclient/-/blob/master/debian/examples), which causes them to be installed in `/usr/share/doc/ddclient/examples`.

SuperSandro2000:
Thanks for the information! I think we just keep them where they are and move them if we add more than one or two new.

PR #138, opened 2020-05-29
Replace dydns.za.net with dyndns.za.net

rhansen:
dydns.za.net doesn't appear to exist.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -2952,11 +2952,11 @@ o 'concont'
 
 The 'concont' protocol is the protocol used by the content management
 system ConCont's dydns module. This is currently used by the free
-dynamic DNS service offered by Tyrmida at www.dydns.za.net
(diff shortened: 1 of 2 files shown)


ddclient
```diff
@@ -2951,11 +2951,11 @@ o 'concont'
 
 The 'concont' protocol is the protocol used by the content management
 system ConCont's dydns module. This is currently used by the free
```
SuperSandro2000, line 2953: Does this need to change, too?

rhansen: It's unclear. If you look at the URL you'll see `dydns` in there, too. That makes me think there used to be a software product called ConCont that did actually have something spelled `dydns`. I can't find anything about ConCont, however.

Perhaps it's time to delete the `concont` protocol?

SuperSandro2000: Maybe the comment needs an update. I am not sure what we should do here.

PR #140, opened 2020-05-30
Fix here doc interpolation

SuperSandro2000:
Right now the -help produces this:
```
./ddclient --help | grep conf
  -file path            : load configuration information from 'path' (default: /etc/ddclient/ddclient.conf).
    -use=smc-barricade-7004vbr  : obtain IP from SMC Barricade FW (7004VBR model config) at the -fw {address}.
    -use=smc-barricade-alt      : obtain IP from SMC Barricade FW (alternate config) at the -fw {address}.
  -if-skip pattern      : skip any IP addresses before 'pattern' in the output of ifconfig {if}.
The configuration file, ${program}.conf, can be used to define the
See the sample-${program}.conf file for further examples.
Example ${program}.conf file entries:
Example ${program}.conf file entries:
```

This PR fixes here doc interpolation that it again looks like:
```
 ./ddclient --help | grep conf
  -file path            : load configuration information from 'path' (default: /etc/ddclient/ddclient.conf).
    -use=smc-barricade-7004vbr  : obtain IP from SMC Barricade FW (7004VBR model config) at the -fw {address}.
    -use=smc-barricade-alt      : obtain IP from SMC Barricade FW (alternate config) at the -fw {address}.
  -if-skip pattern      : skip any IP addresses before 'pattern' in the output of ifconfig {if}.
The configuration file, ddclient.conf, can be used to define the
See the sample-ddclient.conf file for further examples.
Example ddclient.conf file entries:
Example ddclient.conf file entries:
Example ddclient.conf file entries:
```

I hope Emacs works well with this. If not we need to come up with another easy solution or all emacs user need to find a way to fix it.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -2351,7 +2351,7 @@ sub nic_examples {
             $separator = "\n";
         }
     }
-    my $intro = <<'EoEXAMPLE';
+    my $intro = <<"EoEXAMPLE";
 == CONFIGURING ${program}
 
 The configuration file, ${program}.conf, can be used to define the
@@ -2530,7 +2530,7 @@ sub header_ok {
 ## nic_dyndns1_examples
 ######################################################################
 sub nic_dyndns1_examples {
-    return <<'EoEXAMPLE';
+    return <<"EoEXAMPLE";
 o 'dyndns1'
 
 The 'dyndns1' protocol is a deprecated protocol used by the free dynamic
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Fixed those three also. Didn't amend the commit to make review easier.

PR #148, opened 2020-06-02
Fix `CONF` in sample init script for Ubuntu

rhansen:
Both Ubuntu and Debian patch ddclient so that `$etc` is `/etc`, not `/etc/ddclient`.


rhansen:
Actually this should probably be left alone in case someone decides to install ddclient manually, not via an official package.

SuperSandro2000:
I threw my scripts to get my IPs in this directory. Not sure where they Debian packages expects me to put it.

rhansen:
Good point. I guess not many people write scripts to get the IP address; they probably just use `web` or `if`. If I needed to write a script for my Ubuntu system I would probably put it in /usr/local/bin or something.

PR #150, opened 2020-06-03
Use a longer default interval if run as ddclientd

### LINKED ISSUE #149: Require daemonization with a non-zero interval
Command line arguments currently override settings in the config file, so there is no way to pass an argument to ddclient that means "daemonize with an interval of 10m unless the config file wants a different non-zero interval." This makes it difficult to write a robust init script (see the [tricky config file parsing in the example init script](https://github.com/ddclient/ddclient/blob/5cd21f7c7eafd2351940c7b0d01230160e4fa3ef/sample-etc_rc.d_init.d_ddclient#L37-L49)) or systemd service unit config.

One way around this is to run ddclient with `$0` set to something like `ddclientd`. If `$0` ends with the letter `d` there is [special logic that sets `-daemon=1m` by default](https://github.com/ddclient/ddclient/blob/5cd21f7c7eafd2351940c7b0d01230160e4fa3ef/ddclient#L42-L45). I'm not a fan of this magical behavior because it's not discoverable, it's not documented, users are likely to be confused by the `d` that appears in `ps`, and there's no way to change the 1m to something else (other than via the config file).

Can we do better? What do you think about these possibilities:

  * Add special `-default-foo=bar` flags that act like `-foo=bar` except they are overridden by the config file. If we did this, the init script could simply pass `-default-daemon=10m`. We can automatically create these extra flags inside `process_args`. A `daemon=0` line in the config file would thwart `-default-daemon`, which isn't ideal but probably not a problem in practice.
  * Add a new `-require-daemon` flag that causes ddclient to error out if the config file (or other flags) does not set daemon to a positive value.
  * Invert the priority: the config file overrides the flags, not the other way around.
  * Other ideas?

rhansen:
One minute is too short for a default value.

Closes #149

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -41,8 +41,7 @@ my $last_msgs = '';
 
 ## If run as *d (e.g., ddclientd) then daemonize by default (but allow
 ## flags and options to override).
-my $daemon_min = interval('60s');
-my $daemon_default = ($programd =~ /d$/) ? $daemon_min : 0;
+my $daemon_default = ($programd =~ /d$/) ? interval('5m') : 0;
 
 use vars qw($file $lineno);
 local $file   = '';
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Addresses is no magic word to close the issue ;).

rhansen:
> Addresses is no magic word to close the issue ;).

Yup, that was intentional. :smile:  I want to keep the issue open until I modify `-daemon` to accept a bool.

PR #153, opened 2020-06-05
Prefer `ip` over `if` over `web` when inferring `use`

rhansen:
If the user passed `-ip` they almost certainly want to use it, even if they also passed `-if` and `-web`.

Similarly, if the user passed `-if` they almost certainly want to use it even if they also passed `-web`.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -21,6 +21,9 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
     minute.
   * The `pid` option is now ignored when ddclient is not daemonized.
   * ddclient now gracefully exits when interrupted by Ctrl-C.
+  * The way ddclient chooses the default for the `use` option has changed.
+    Rather than rely on the default, users should explicitly set the `use`
+    option.
 
 ## 2020-01-08 v3.9.1
 
--- a/ddclient
+++ b/ddclient
@@ -1291,9 +1291,11 @@ sub init_config {
     $opt{'quiet'} = 0 if opt('verbose');
 
     ## infer the IP strategy if possible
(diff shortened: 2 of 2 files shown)


SuperSandro2000:
How and when do we wanna do the release notes? I think this should be mentioned so that no one is surprised by sudden changes.

rhansen:
> How and when do we wanna do the release notes? I think this should be mentioned so that no one is surprised by sudden changes.

Agreed. Some of the previous changes should be mentioned too. Let's hold off on submitting this until we have a release notes doc.

rhansen:
I made PR #156 for documenting release notes and updated this PR to add a note. I'll remove the "draft" status of this PR after PR #156 is merged.

SuperSandro2000:
Merged it. Thanks for the work!

ChangeLog.md
```diff
@@ -21,6 +21,9 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
     minute.
   * The `pid` option is now ignored when ddclient is not daemonized.
   * ddclient now gracefully exits when interrupted by Ctrl-C.
+  * The way ddclient chooses the default for the `use` option has changed.
```
SuperSandro2000, line 24: I think it would be nice to mention how it is chosen now.

rhansen: I intentionally omitted that because I want users to explicitly set `use`. I view this feature as "the config is broken, but let's try to do something useful anyway". Should we change the documentation for `use` to make this behavior official?

SuperSandro2000: Yeah, you're right they should explicitly set it.

PR #155, opened 2020-06-07
Fix regex searching for Content-Type header

rhansen:
* Add the `m` modifier because the `$headers` variable can contain multiple headers. If `$headers` contains a Content-Type header but it is the second or later header then the regex won't match without `m`.
  * Add the `i` modifier because RFC 7230 says that header field names are case-insensitive.
  * Don't require a space after the colon because RFC 7230 says that the space is optional.


rhansen:
I updated the commit message; please take another look.

PR #156, opened 2020-06-07
Start a NEWS document for notable changes

rhansen:


rhansen:
I reverted the merge. I would reopen this pull request but it looks like I don't have permission, so I'll open a new pull request so we can discuss further.

PR #158, opened 2020-06-09
Convert RELEASENOTE to a history of notable changes

rhansen:
PR #156 was reverted due to a post-merge comment. This is a re-do of that PR so that we can discuss further.


rhansen:
Currently there are three files describing changes:

  * `ChangeLog`: Contains a detailed history of every svn commit.
  * `Changelog.old`: Per-release summary of important changes for v3.7.3 and older.
  * `RELEASENOTE`: Summary of important changes in only the most recent release.

Most open source projects follow this convention:

  * `ChangeLog`: Detailed history of every commit. This file is becoming more and more rare because it's tedious to maintain and the same information can be gleaned from reviewing the version control system's log.
  * `NEWS`: Per-release summary of important changes. This contains the most important points that users will want to see. It is a critical resource for package maintainers, who use it to determine if there compatibility-breaking changes or updated dependencies.

This PR currently converts `RELEASENOTE` into `NEWS` (backfilling old versions) but leaves `ChangeLog` and `Changelog.old` alone. @wimpunk: You mentioned that you didn't like the name `NEWS`. What is your preference?

SuperSandro2000:
Changelog.old has a similar format to news.md. I think it should be merged and then deleted.

wimpunk:
Some historical information: `Changelog.old` was generated by some svn commando, at that time it was just called `Changelog` (or maybe `ChangeLog`, not completely sure).  At some time I started using git and I couldn't generated the same kind of changelog so I decided to rename it to `Changelog.old` and use `Changelog` for the git history.   Unfortunatly it's currently impossible by running some thing like `git log --oneline --since v3.9.0` to see a lot of useful information so I mostly generate the contents and change it where needed.  I put the latest version in the `RELEASENOTE` because it used to be picked up while releasing using sf.net.  I think I have to copy the contents to the webinterface while releasing on github.
So for historical reasons I would like to keep the Changelog file but I'm open for good arguments.  We need at least a place where we find a good overview about what has changed.  Currently `git log --oneline --since=v3.9.1` shows a lot of info but it isn't very useful for users.  Thinking of it, if we don't fill the `Changelog` with the git history, we could start using the suggested `NEWS` file but we may not forget to fill it with the info needed.  Things I would like to see in the `NEWS` file:
- new, updated or removed providers
- new, updated or removed routers
- code cleanup: I've always wanted to do this but it was hard to do when using svn because the patches wouldn't apply anymore.  So thanks for defining a code style and cleanup the code.
- important changes like dependencies
There could be more items but these are the ones I currently think about.

rhansen:
> Unfortunatly it's currently impossible by running some thing like `git log --oneline --since v3.9.0` to see a lot of useful information

I'm curious about why you say that. Is it because there are lots of merge commits? If so, you can pass `--no-merges`. Or is it because there are a lot more cleanup and minor tweak commits that aren't newsworthy?

> so I mostly generate the contents and change it where needed.

I think that's the only thing you can do. Many projects have tried to get devs to add a useful "Release-Note:" pseudo-header in commit messages for automatic extraction by a script. That approach usually fails: devs inevitably forget to add that pseudo-header, they don't word it appropriately for user consumption, they add it for trivial stuff, etc.  Someone usually ends up having to manually edit the release notes anyway.

Usually I try to add an entry to the release notes doc in the same commit that makes the newsworthy change. It's similar to the "Release-Note:" pseudo-header approach but has some advantages: (1) there's no need to run a script to generate the release notes, (2) people can easily see what's going to be in the next release, and (3) there's something about the psychology of editing the release notes directly that causes devs to do a better job at wording the entry for public consumption. The major downside is it increases the probability of a merge conflict. And someone will still have to review the notes before release anyway. Hopefully that's an easier task if the notes are updated along the way.

> So for historical reasons I would like to keep the Changelog file but I'm open for good arguments. We need at least a place where we find a good overview about what has changed.

I don't mind calling it "ChangeLog" if that's your preference. What do you think about this:
  1. Rename this PR's `NEWS.md` to `ChangeLog.md`. (I think we want the `.md` extension so that it renders nicely in GitHub.)
  2. Move the existing `ChangeLog` contents to `ChangeLog.md`, but group the entries by release version instead of date.
  3. Move the `Changelog.old` contents as-is to `ChangeLog.md`.

> Thinking of it, if we don't fill the `Changelog` with the git history, we could start using the suggested `NEWS` file but we may not forget to fill it with the info needed.

Do you mean "may forget"? That's a risk, but I think it's a minor one: When it's time to cut a new release we'll remember to update the release notes at that time, and if we still forget something we can always go back and add it when we do remember.

> Things I would like to see in the `NEWS` file:
> 
>   * new, updated or removed providers
>   * new, updated or removed routers
>   * code cleanup: I've always wanted to do this but it was hard to do when using svn because the patches wouldn't apply anymore.  So thanks for defining a code style and cleanup the code.
>   * important changes like dependencies

:+1:

Regarding code cleanup, I usually just add a single bullet that says something like "internal cleanup" (unless it's a major refactor) because it doesn't affect users much and we want a good signal to noise ratio.

SuperSandro2000:
> I'm curious about why you say that. Is it because there are lots of merge commits? If so, you can pass --no-merges. Or is it because there are a lot more cleanup and minor tweak commits that aren't newsworthy?

If so those can be ignored with: ``git log --oneline --since v3.9.0 --no-merges``

rhansen:
I renamed `NEWS.md` to `ChangeLog.md` and merged in the contents of `ChangeLog` and `Changelog.old`. What do you think about it now?

rhansen:
I have 4 branches that are based on this PR, so unless anyone vetos I'm going to merge this later tonight.

Issue #146, opened 2020-06-02
Create a man page


rhansen:
The output of `-help` is quite unwieldy. I think it would be good to move the examples to a man page to drastically shrink the amount of help text. The man page would also be a good place to clarify [the config file format](https://github.com/ddclient/ddclient/blob/5cd21f7c7eafd2351940c7b0d01230160e4fa3ef/ddclient#L1128-L1157).

wimpunk:
I think we should also add the contents of the website to the repository.  It's bit unpractical to maintain two repositories which contain related material.
The website is based on markdown files which are rendered by the github infrastructure.  You can find the code at https://github.com/ddclient/ddclient.github.io.  As far as I remember we could add a directory `doc` which can be used as source for the github pages.  I have to check this, I'm not completely sure about the directory name.
Documenting the code was a good thing to use but I'm not sure we have to keep it that way.  I believe we need one documentation source which can generate the output we need.  I've always been dreaming about integrating it in ddclient so we could run `ddclient --help --web` to generate the website an drun `ddclient --help --man` to create the man pages but this is probably to complex to setup.
I also hope to keep the documentation together.  I'm not in favor of shrinking the amount of help text, I would prefer the possibility to select which part of the documentation you want to read.  But again, it could make the help feature to complex.
So if you make any changes to the help, try to keep all documentation together.  It may be all together in one directory, that's good enough.   I just don't want to have to search on multiple places to find all the information.
Anyone from @ddclient/ddclientcore who wants to share his vision?

SuperSandro2000:
The idea was to move stuff from --help to the man page to minimize the help. Currently it is very long and may be confusing.

wimpunk:
I know the result is pretty long. But it should also be easy readable without the need to install a man page handler.  Maybe I'm just getting old but I prefer a human readable document above a man page. Adding the info to ddclient makes it easier to update.  Currently there's already a difference between what ddclient can do and what's in the documentation.  I've always considered `ddclient --help` as the only source of truth.  So if we want to reduce the help-feature, we have to find a way to easily maintain the documentation and generate the contents of the website and man pages from that source.

SuperSandro2000:
If you type git --help you are automatically show the man page or just the arguments. Also I expect everyone to have man installed.

I never read --help and just searched through the script file. For me it is way to long to read and my less does not support mouse scroll so it is just inconvenient.

There are tools which can convert man files to markdown or HTML (groff) and the other way (roff).

[...]

Issue #109, opened 2020-04-13
ddclient runs as root by default.


tts3c:
The ddclient service runs as root by default. The default cron config also suggest to run the client as root. 
$ ps -aux | grep ddclient
root      655098  0.0  0.0  21100 14296 ?        S    10:43   0:00 ddclient - sleeping for 300 seconds

The client can also run as a user, provided the files permissions are set accordingly. It should not run as root by default, specially if it is running as a daemon.

wimpunk:
Actually, it depends on the distrubution you're using or the configuration you made.  So if you installed ddclient using the package of your distribution you should better ask to maintainer of the package to change the rights.

rhansen:
I think this is in-scope for the ddclient project, so I'm removing the wontfix label.

In general, daemons should write the PID file as root then drop privileges. (The PID file should be owned by root so that an attacker that compromises ddclient can't cause the init system to kill another process by writing the other process's PID in the ddclient PID file. systemd has some safety checks to avoid this, but it's still good practice.) With ddclient as it is now, the only way to write a PID file as root and then drop privileges is to launch ddclient via a helper utility that daemonizes, writes the PID file, drops privileges, then execs ddclient with `-foreground`. If distributions are expected to do this then ddclient's daemonization functionality might as well be removed.

taimoortalpur:
For more info. I am using arch linux and have also reported it to the distribution. 
https://bugs.archlinux.org/task/66893?string=ddclient

wimpunk:
@rhansen We can suggest users to use another user but we will have to think about how to fix the cache location.  I really don't believe this is something which has to fixed in the code, although we could check which user is running ddclient and switch to another cache location but I don't think there are any recommentations about cache locations for non-root users.

rhansen:
I don't think we need to do anything. As long as the cache file isn't opened until after privileges are dropped, users can simply chown `/var/cache/ddclient` to the unprivileged user and everything should just work. (As far as I can tell, nothing in [Debian policy](https://www.debian.org/doc/debian-policy/) prohibits changing the owner of `/var/cache/<package>` to an unprivileged user.)

wimpunk:
I was thinking about a multiuser system where user A runs ddclient and user B runs ddclient but if you think about running it as a demon, it would make sense by just changing the ownership of the cache file.  Maybe we could add some info about the readme on how to do it, although that's pretty obvious for most of the users.

SuperSandro2000:
[...]

PR #161, opened 2020-06-09
Build system and unit test infrastructure

### LINKED ISSUE #147: Unit test infrastructure
It would be nice if ddclient had some infrastructure to make it easy to write and run unit tests.

rhansen:
See the individual commits for details.

Addresses #146 
Fixes #147


autogen
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+cd "${0%/*}" || exit 1
+mkdir -p m4 build-aux || exit 1
```
SuperSandro2000, line 4: You could replace ``|| exit 1`` with ``set -e``.

rhansen: True. In general I prefer to explicitly exit (instead of `set -e`) for a few reasons:

  * If I'm in the habit of *always* doing `|| exit 1` then I'm more likely to notice when it's missing. If I change my habit to rely on `set -e` then I'm less likely to notice when `set -e` is missing in a script.
  * The POSIX specification for `set -e` is not very clear, and different shells do different things in corner cases, such as `eval false` and `x=$(false)`.
  * In more complicated scripts I like to add error messages:

    ```shell
    pecho() { printf %s\\n "$*"; }
    log() { pecho "$@"; }
    error() { log "ERROR: $@" >&2; }
    fatal() { error "$@"; exit 1; }

    foo bar || fatal "an error message goes here"
    ```

SuperSandro2000: I always set ``set -eou pipefail``.

SuperSandro2000:
You sure this is not a little bit to much?

rhansen:
It is quite heavyweight, but I wonder if there's a better alternative. I'm not familiar enough with cmake to know if it would be any lighter. Maybe there's something in the Perl ecosystem, but I didn't see one.

We could just write our own Makefile, but it's difficult to do that properly with good multi-platform support. Also, hand-crafted Makefiles tend to get complicated and ugly as a project evolves.

autogen
rhansen: Unfortunately `-o pipefail` is not in [the POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#set) (it's a bashism).

ChangeLog.md
```diff
@@ -8,6 +8,10 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
 ### New features
 
   * Added support for OVH DynHost.
+  * Build system to make it easier for distributions to
+    package ddclient:
+
+        ./configure && make && make check && make install
```
SuperSandro2000, line 14:
```suggestion
        ./autogen && ./configure && make && make check && make install
```

Makefile.am
```diff
@@ -0,0 +1,37 @@
...
+	sample-get-ip-from-fritzbox
+
+LOG_DRIVER = env AM_TAP_AWK='$(AWK)' $(SHELL) \
+	$(top_srcdir)/build-aux/tap-driver.sh
+TEST_EXTENSIONS = .pl
+PL_LOG_DRIVER = $(LOG_DRIVER)
+PL_LOG_COMPILER = $(PERL)
+AM_PL_LOG_FLAGS = -Mstrict -w -I'$(abs_top_srcdir)'
+TESTS = \
+	t/load.pl
```
SuperSandro2000, line 36: I think this should be replaced with something that just loads all files in ``t/``.

build-aux/tap-driver.sh
```diff
@@ -0,0 +1,651 @@
+#! /bin/sh
+# Copyright (C) 2011-2020 Free Software Foundation, Inc.
```
SuperSandro2000, line 2:
```suggestion
# Copyright (C) 2011-2018 Free Software Foundation, Inc.
```
I think my system already has that script file installed and I got this diff. I am not sure if we want to apply this diff or rather put it into the gitignore file.

ChangeLog.md
rhansen: Done.

Makefile.am
rhansen: That would be nice, but Automake intentionally does not support globs.

build-aux/tap-driver.sh
rhansen: Unfortunately adding it to the gitignore file won't work because the file is tracked (Git ignores .gitignore for tracked files).

This is one of the reasons I don't like to check in generated files. Whenever you run `autogen` you will see diffs like this unless the version of Automake you installed on your machine matches the version of Automake that provided this version of `tap-driver.sh`.

This file is checked in because Automake 1.11 and older doesn't come with it, and RHEL/CentOS 6 has Automake 1.11. RHEL/CentOS 6 reach end of life on November 30th, 2020, at which point we can bump the minimum required Automake version and delete this file from our repository.

rhansen:
> I think the readme needs some updates for this.

I will add something once we've committed to this.

> So I think if we are going more towards a package with more complex dependencies and setup this is great but if we stay just a script to execute this is too much. I am not sure where we are heading right now but I don't really want to debug dependency problems.

Here's where I think ddclient is going in the near future:
  * Move documentation to a separate file and "build" it into the ddclient script (#146). How I would expect this to look:
      * The `ddclient` file would be renamed to something like `ddclient.template`. The documentation that currently exists in `ddclient` would be replaced with a token for easy substitution (e.g., `@DOCUMENTATION_GOES_HERE@`). Ideally the template file would still be directly executable without building anything so that it is still easy to develop new features or fix bugs.
      * Autoconf would check for an acceptable XML or Markdown parser and error out if it isn't found (or activate conditional logic that omits the documentation).
      * Makefile.am would get a few new rules:
           * A rule to generate the man page from the XML or Markdown.
           * A rule to install the man page to the appropriate directory.
           * A rule to generate HTML from the XML or Markdown (for https://ddclient.net).
           * A rule to generate plain text from the XML or Markdown.
           * A rule to build the `ddclient` script by taking the template file and substituting in the plain text documentation.
  * A build rule to replace the `#!/usr/bin/perl` shebang with the path discovered by Autoconf.
  * New dependency on curl. This would either be a dependency on the `curl` executable or on a Perl module that uses `libcurl`. Autoconf would check for the dependency and error out if not found.
  * New dependency on Proc::Daemon. Proper daemonization is not trivial; there are several things ddclient does wrong. Using Proc::Daemon would eliminate a race condition with the creation of the PID file, and it would make it easy for us to drop root privileges (#109).
  * Many new and exotic dependencies that will be used only for unit tests. These dependencies would include things like a dummy HTTP server. If one of these dependencies is not found by Autoconf, a warning will be printed but it won't fail—the relevant unit tests would instead be skipped.

Overall, having some sort of build system gives us a lot more flexibility without increasing the burden on our users (including packagers) too much, though there is a significant complexity cost. I think the pros outweigh the cons. The only thing I'm not sure about is whether Autotools is the right choice here (vs. cmake or something else).

build-aux/tap-driver.sh
SuperSandro2000: Git should ignore those changes if you commit the gitignore change.

dkerr64:
> * New dependency on curl. This would either be a dependency on the `curl` executable or on a Perl module that uses `libcurl`. Autoconf would check for the dependency and error out if not found.
> * New dependency on Proc::Daemon. Proper daemonization is not trivial; there are several things ddclient does wrong. Using Proc::Daemon would eliminate a race condition with the creation of the PID file, and it would make it easy for us to drop root privileges (#109).

We need to minimize new dependencies on Perl modules, especially ones not installed by default.  So for curl we should use the curl executable command and not require pulling in another Perl module.  And because curl is more robust than our current implementation we should probably default to using it if the executable exists.

SuperSandro2000:
> especially ones not installed by default

As long as they are available in the package repos I see no reason not to use them. Even if we use one that is not available on centos7 but still all other systems I would rather use it than coding advanced stuff ourselves.

Proc::Daemon is available in Debian https://packages.debian.org/buster/libproc-daemon-perl so feel free to use it from my side.

SuperSandro2000:
How is the status on this? A lot of other PRs are depending on this and I have the feeling that it would easen a lot of tasks for you and the others if we would finalize this. Do you need help testing or some decisions to make?

rhansen:
> We need to minimize new dependencies on Perl modules

With a build system in place we can use https://metacpan.org/pod/PAR to combine the script with its pure Perl module dependencies into a single executable. Of course the dependencies would have to be available at build time, but if we were concerned about their availability on a platform we wish to support we could copy them into the ddclient Git repository.

> How is the status on this?

I think it can be merged. I hadn't marked this as ready for review yet because I kept finding little issues while I was working on the geturl unit tests. But I think it's in shape now.

dkerr64:
> With a build system in place we can use https://metacpan.org/pod/pp to combine the script with its pure Perl module dependencies into a single executable. Of course the dependencies would have to be available at build time, but if we were concerned about their availability on a platform we wish to support we could copy them into the ddclient Git repository.

Will this work in a cross compile environment like you would find when creating a system with Buildroot?  What if target architecture is different (e.g. build on x86 run on ARM).

rhansen:
Good question, and relevant for embedded platforms like consumer routers. Looks like the answer is [no, cross-compiling is not supported](https://metacpan.org/pod/PAR::FAQ#The-Perl-Packager-scripts-says-that-it-can-create-executable-that-runs-in-same-OS.-Can-I-use-it-to-create-Win32-binary-with-linux-machine?-Or-what-should-I-use-to-create-Win32-executable-binary-on-linux-from-my-script?). Maybe [B::C](https://metacpan.org/pod/B::C) would be a viable approach? Or you could run `pp` inside QEMU, though that would probably take a lot of effort to set up.

rhansen:
> > How is the status on this?
> 
> I think it can be merged.

Actually there are a couple of things I still need to do:
  * Update README.md
  * I'm getting an error on Debian Sid that I need to figure out.

SuperSandro2000:
> Or you could run pp inside QEMU, though that would probably take a lot of effort to set up.

You can quite easily set up binfmt multiarch in Docker container but we could not easily test it if it is true multiarch and not some x86 left behind in it without actual hardware.

SuperSandro2000:
> Update README.md

A quick list of packages I needed to install to get autoconf running in docker:

``apt install git autoconf autoconf-archive``

This should probably go into the readme in some form.

Notes:

Without autoconf-archive I got the following cryptic error:

```
configure.ac:23: error: possibly undefined macro: AC_MSG_ERROR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
```


and this is the full list of all packages I needed in the end:

```
apt install git autoconf autoconf-archive libdata-validate-ip-perl make
```


The tests are failing for me with:

```
# make check
make  check-TESTS
make[1]: Entering directory '/ddclient'
make[2]: Entering directory '/ddclient'
FAIL: t/version.pl 1 - require 'ddclient'
ERROR: t/version.pl - Bail out! ddclient did not return a true value at ./t/version.pl line 4.
============================================================================
Testsuite summary for ddclient 3.9.1
============================================================================
# TOTAL: 2
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 1
============================================================================
See ./test-suite.log
============================================================================
make[2]: *** [Makefile:549: test-suite.log] Error 1
make[2]: Leaving directory '/ddclient'
make[1]: *** [Makefile:657: check-TESTS] Error 2
make[1]: Leaving directory '/ddclient'
make: *** [Makefile:855: check-am] Error 2

```

and when I just run ddclient I get this pile of errors and warnings:

```
./ddclient
WARNING:  file /etc/ddclient/ddclient.conf: Cannot open file '/etc/ddclient/ddclient.conf'. (No such file or directory)
stat() on closed filehandle FD at ./ddclient line 1178.
Use of uninitialized value $mode in bitwise and (&) at ./ddclient line 1179.
readline() on closed filehandle FD at ./ddclient line 1190.
WARNING:  file /etc/ddclient/ddclient.conf: Cannot open file '/etc/ddclient/ddclient.conf'. (No such file or directory)
stat() on closed filehandle FD at ./ddclient line 1178.
Use of uninitialized value $mode in bitwise and (&) at ./ddclient line 1179.
readline() on closed filehandle FD at ./ddclient line 1190.
```

We should error if we can't read the config file instead of continuing.


Also ``make install`` should probably place a default conf at ``/etc/ddclient/ddlclient.conf``.


I hope this feedback helps you debugging your problem and getting this PR done quicker.

dkerr64:
> We should error if we can't read the config file instead of continuing.

Not sure I agree.  When testing ddclient I sometimes use -query which doesn't require a conf file or any settings on command line, but does test all the default ways of getting an IP address.  There may be other parameters where proceeding without a conf file makes sense.

dkerr64:
Docker / QEMU. Really? You're kidding right? Please no.  For ddclient to require either to run on certain systems (lets say to ensure required packages are available) is a non-starter.  The overhead of either is many multiples of the size of ddclient and is a non-starter.

dkerr64:
> and this is the full list of all packages I needed in the end:
> 
> ```
> apt install git autoconf autoconf-archive libdata-validate-ip-perl make
> ```
I thought we got rid of Data::Validate::IP requirement.  Maybe still in a PR to get merged?

SuperSandro2000:
> I thought we got rid of Data::Validate::IP requirement. Maybe still in a PR to get merged?

Good catch.

> Not sure I agree. When testing ddclient I sometimes use -query which doesn't require a conf file or any settings on command line, but does test all the default ways of getting an IP address. There may be other parameters where proceeding without a conf file makes sense.

Or at least set those vars to some default and handle the case cleanly.

configure.ac
```diff
@@ -0,0 +1,29 @@
...
+# above AC_REQUIRE_AUX_FILE line from causing configure to complain
+# about a mising file if the user has Automake 1.11.)
+AM_INIT_AUTOMAKE([1.11 -Wall -Werror foreign subdir-objects])
+AM_SILENT_RULES
+
+AX_WITH_PROG([PERL], perl)
+AX_PROG_PERL_VERSION([5.10.1], [],
+  [AC_MSG_ERROR([Perl 5.10.1 or newer not found])])
+m4_foreach_w([_module], [
+    Data::Validate::IP
```
SuperSandro2000, line 21: This needs an update after https://github.com/ddclient/ddclient/pull/178 got merged.

dkerr64:
> > Not sure I agree. When testing ddclient I sometimes use -query which doesn't require a conf file or any settings on command line, but does test all the default ways of getting an IP address. There may be other parameters where proceeding without a conf file makes sense.
> 
> Or at least set those vars to some default and handle the case cleanly.
Yes, I agree it should be handled cleanly.

configure.ac
rhansen: Done.

rhansen:
> > Update README.md
> 
> A quick list of packages I needed to install to get autoconf running in docker:
> 
> `apt install git autoconf autoconf-archive`

`autoconf-archive` should no longer be required.

> 
> This should probably go into the readme in some form.

Done.

> 
> The tests are failing for me with:

Should be fixed now.

> 
> Also `make install` should probably place a default conf at `/etc/ddclient/ddlclient.conf`.

Another reason this PR is still draft is the resulting ddclient script doesn't honor `./configure --sysconfdir=/etc/ddclient`. I need to fix that before I can install a default config.

> 
> I hope this feedback helps you debugging your problem and getting this PR done quicker.

Yes, thank you!

> > We should error if we can't read the config file instead of continuing.
> 
> Not sure I agree.

Would one of you mind opening a new issue for that so we can discuss further?

> Docker / QEMU. Really? You're kidding right? Please no. For ddclient to require either to run on certain systems (lets say to ensure required packages are available) is a non-starter.

I was talking about QEMU as a way to cross-build a ddclient binary, not run QEMU itself on the target platform.

SuperSandro2000:
> Would one of you mind opening a new issue for that so we can discuss further?



I opened #188

rhansen:
I made some significant changes. Highlights:
  * `ddclient` was renamed to `ddclient.template` so that we can substitute user-provided alternatives to `/var/cache`, `/etc/ddclient`, and `/usr/bin/perl`.
  * `sample-etc_ddclient.conf` was renamed to `ddclient.conf.template` so that we can substitute a user-provided alternative to `/var/run` and install it in `/etc`.
  * `make install` now installs the sample `ddclient.conf` and creates `/var/cache/ddclient`.

The script should still be runnable from Git, except you will have to invoke it as `perl ./ddclient.template` rather than just `./ddclient.template` (because the shebang line is now a substitution token).

rhansen:
Unless anyone objects, I'm going to merge this tomorrow.

SuperSandro2000:
If you resolve the merge conflicts I will merge it.

PR #163, opened 2020-06-10
Design doc for IPv6 support

rhansen:
Before anyone spends too much time implementing IPv6 support, I want to make sure we all agree on what IPv6 support should look like. This PR contains my attempt at describing what we might want. All interested parties are encouraged to review this and propose changes (or voice agreement).

Addresses #172

### DIFF
--- a/docs/ipv6-design-doc.md
+++ b/docs/ipv6-design-doc.md
@@ -1,0 +1,337 @@
+# Design Doc: IPv6 Support
+
+Author: [@rhansen](https://github.com/rhansen/)\
+Date: 2020-06-09\
+Signed off by:
+[@SuperSandro2000](https://github.com/SuperSandro2000/)
+
+## Objective
+
+Add full IPv6 support to ddclient, including support for dual-stack
+systems.
+
+## Background
+
+ddclient's current IPv6 support is limited:
+
+  * Users can update either an IPv6 record or an IPv4 record for a
+    host, not both.
+  * If SSL is used for an HTTP request, IPv6 will be used if the
+    remote host has a AAAA record, even if the user would rather use
+    IPv4. This breaks `use=web` for IPv4 if the `web` URL's host has a
+    AAAA record.
+  * The `use=if` method only works if the user sets `if-skip` to
+    something that skips over all IPv4 addresses in the output of
+    `ifconfig` (or `ip`). If the output contains an IPv4 address after
+    the IPv6 address then `use=if` cannot be used for IPv6.
+  * There is no support for falling back to IPv4 if an IPv6 connection
+    fails.
+  * `use=if` does not filter out locally scoped or temporary IPv6
+    addresses.
+
+Some attempts have been made to add more robust IPv6 support:
+
+  * Debian's ddclient package applies a
+    [patch](https://salsa.debian.org/debian/ddclient/-/blob/67a138aa3d98d70f01766123f58ef40e98693fd4/debian/patches/usev6.diff)
+    that adds a new `usev6` option. The `usev6` option can be set to
+    `ip` or `if`, but not any of the other strategies currently
+    available for the `use` option (`web`, `cmd`, `fw`, `cisco`,
+    `cisco-asa`). When set to `ip` or `if`, only IPv6 addresses are
+    considered; IPv4 addresses are ignored. The patch does not change
+    the behavior of the `use` option, so `use=web` or `use=cmd` can be
+    used for IPv6 if pointed at something that only outputs an IPv6
+    address.
+  * [ddclient-curl](https://github.com/astlinux-project/ddclient-curl)
+    is a fork of ddclient that uses curl as the HTTP client (instead
+    of ddclient's own homemade client) for more robust IPv6 support.
+  * PR #40 is perhaps the most comprehensive attempt at adding full
+    IPv6 support, but it was never merged and has since
+    bit-rotted. There is renewed effort to rebase the changes and get
+    them merged in. PR #40 adds new options and changes some existing
+    options. The approach taken is to completely isolate IPv4 address
+    detection from IPv6 address detection and require the update
+    protocol callbacks to handle each type of address appropriately.
+
+## Requirements
+
+  * The mechanism for determining the current IPv4 address (the `use`
+    option) must be independently configurable from the mechanism used
+    to determine the current IPv6 address.
+  * The user must be able to disable IPv4 address updates without
+    affecting IPv6 updates.
+  * The user must be able to disable IPv6 address updates without
+    affecting IPv4 updates.
+  * If HTTP polling is used for both IPv4 and IPv6 address discovery,
+    the URL used to determine the IPv4 address (the `web` option) must
+    be independently configurable from the URL used to determine the
+    IPv6 address.
+  * The use of IPv4 or IPv6 to update a record must be independent of
+    the type of record being updated (IPv4 or IPv6).
+  * The callback for the update protocol must be given both addresses,
+    even if only one of the two addresses has changed.
+  * The callback for the update protocol must be told which addresses
+    have changed.
+  * There must be IPv6 equivalents to `use=ip`, `use=if`, `use=web`,
+    and `use=cmd`. For the IPv6 equivalent to `use=if`, it is
+    acceptable to ignore non-global and temporary addresses (the user
+    can always use the IPv6 equivalent to `use=cmd` to get non-global
+    or temporary addresses).
+  * Existing support for updating IPv6 records must not be lost.
+  * Some dynamic DNS service providers use separate credentials for
+    the IPv4 and IPv6 records. These providers must be supported,
+    either by accepting both sets of credentials in a single host's
+    configuration or by allowing the user to specify the same host
+    twice, once for IPv4 and once for IPv6.
+
+### Nice-to-Haves
+
+  * The user should be able to force the update protocol to use IPv4
+    or IPv6.
+  * Unless configured otherwise, ddclient should first attempt to
+    update via IPv6 and fall back to IPv4 if the IPv6 connection
+    fails. This behavior can be added later; for now it is acceptable
+    to keep the current behavior (use IPv6 without IPv4 fallback if
+    there is a AAAA record, use IPv4 if there is no AAAA record).
+  * Full backwards compatibility with existing config files and
+    flags. The trade-offs between migration burden, long-term
+    usability, and code maintenance should be carefully considered.
+  * IPv6 equivalents to `use=fw`, `use=cisco`, and `use=cisco-asa`.
+  * Add IPv6 support in protocol callbacks where IPv6 support is
+    currently missing. (This can be done later.)
+
+## Proposal
+
+### Configuration changes
+
+  * Add new `usev4` and `usev6` settings that are like the current
+    `use` setting except they only apply to IPv4 and IPv6,
+    respectively.
+      * `usev4` can be set to one of the following values: `disabled`,
+        `ipv4`, `webv4`, `fwv4`, `ifv4`, `cmdv4`, `ciscov4`,
+        `cisco-asav4`
+      * `usev6` can be set to one of the following values: `disabled`,
+        `ipv6`, `webv6`, `fwv6`, `ifv6`, `cmdv6`, `ciscov6`,
+        `cisco-asav6`
+  * Add a new `use` strategy: `disabled`.
+  * The `disabled` value for `use`, `usev4`, and `usev6` causes
+    ddclient to act as if it was never set. This is useful for
+    overriding the global value for a particular host.
+  * For compatibility with ddclient-curl, `no` is a deprecated alias
+    of `disabled`.
+  * Add new `ipv4`, `ipv6`, `webv4`, `webv4-skip`, `webv6`,
+    `webv6-skip`, `ifv4`, `ifv6`, `cmdv4`, `cmdv6`, etc. settings that
+    behave like their versionless counterparts except they only apply
+    to IPv4 or IPv6. Deprecate the versionless counterparts, and
+    change their behavior so that they also influence the default
+    value of the versioned options. (Example:  Suppose
+    `usev4=ifv4`. If `ifv4` is not set then `if` is used.)  Special
+    notes:
+      * The value of `ip` will only serve as the default for `ipv4`
+        (or `ipv6`) if it contains an IPv4 (or IPv6) address.
+      * There is currently an `ipv6` boolean setting. To preserve
+        backward compatibility with existing configs, `ipv6` set to a
+        boolean value is ignored (other than a warning).
+      * There is no `ifv4-skip` or `ifv6-skip` because it's ddclient's
+        responsibility to properly parse the output of whatever tool
+        it uses to read the interface's addresses.
+      * For now there is no `cmdv4-skip` or `cmdv6-skip`. Anyone who
+        already knows how to write a regular expression can probably
+        write a wrapper script. These may be added in the future if
+        users request them, especially if it facilitates migration
+        away from the deprecated `cmd-skip` setting.
+      * For `usev6=ifv6`, interfaces are likely to have several IPv6
+        addresses (unlike IPv4). Choosing the "right" IPv6 address is
+        not trivial. Fortunately, we don't have to solve this
+        perfectly right now; we can choose something that mostly
+        works and let user bug reports guide future refinements. For
+        the first iteration, we will try the following:
+          * Ignore addresses that are not global unicast.
+            (Unfortunately, the `ip` command from iproute2 does not
+            provide a way to filter out ULA addresses so we will have
+            to do this ourselves.)
+          * Ignore temporary addresses.
+          * If no addresses remain, log a warning and don't update the
+            IPv6 record.
+          * Otherwise, if one of the remaining addresses matches the
+            previously selected address, continue to use it.
+          * Otherwise, select one arbitrarily.
+  * Deprecate the `use` setting (print a loud warning) but keep its
+    existing semantics with an exception: If there is a conflict with
+    `usev4` or `usev6` then those take priority:
+      * If `use`, `usev4`, and `usev6` are all set then a warning is
+        logged and the `use` setting is ignored.
+      * If `use` and `usev4` are both set and the `use` strategy
+        discovers an IPv4 address that differs from the address
+        discovered by the `usev4` strategy, then the address from
+        `usev4` is used and a warning is logged.
+      * If `use` and `usev6` are both set and the `use` strategy
+        discovers an IPv6 address that differs from the address
+        discovered by the `usev6` strategy, then the address from
+        `usev6` is used and a warning is logged.
+  * If `usev4` (`usev6`) is not set:
+      * If `ipv4` (`usev6`) is set, ddclient acts as if `usev4`
+        (`usev6`) was set to `ipv4` (`ipv6`).
+      * Otherwise, if `ifv4` (`ifv6`) is set, ddclient acts as if
+        `usev4` (`usev6`) was set to `ifv4` (`ifv6`).
+      * Otherwise, if `cmdv4` (`cmdv6`) is set, ddclient acts as if
+        `usev4` (`usev6`) was set to `cmdv4` (`cmdv6`).
+      * Otherwise, if `fwv4` (`fwv6`) is set, ddclient acts as if
+        `usev4` (`usev6`) was set to `fwv4` (`fwv6`).
+      * Otherwise, `usev4` (`usev6`) remains unset.
+  * To support separate credentials for IPv4 vs. IPv6 updates, users
+    can specify the same host multiple times, each time with different
+    options.
+
+### Internal API changes
+
+  * Add two new entries to the `$config{$host}` hash:
+      * `$config{$host}{'wantipv4'}` is set to:
+          * If `usev4` is enabled, the IPv4 address discovered by the
+            `usev4` strategy.
+          * Otherwise, if `use` is enabled and the `use` strategy
+            discovered an IPv4 address, the IPv4 address discovered by
+            the `use` strategy.
+          * Otherwise, `undef`.
+      * `$config{$host}{'wantipv6'}` is set to:
+          * If `usev6` is enabled, the IPv6 address discovered by the
+            `usev6` strategy.
+          * Otherwise, if `use` is enabled and the `use` strategy
+            discovered an IPv6 address, the IPv6 address discovered by
+            the `use` strategy.
+          * Otherwise, `undef`.
+  * Deprecate the existing `$config{$host}{'wantip'}` entry, to be
+    removed after all update protocol callbacks have been updated to
+    use the above new entries. In the meantime, this entry's value
+    depends on which of `use`, `usev4`, and `usev6` is enabled, and
+    what type of IP address is discovered by the `use` strategy (if
+    enabled), according to the following table:
+
+    | `use` | `usev4` | `usev6` | resulting value |
+    | :---: | :---: | :---: | :--- |
+    | ✔(IPv4) | ✖ | ✖ | the IPv4 address discovered by the `use` strategy |
+    | ✔(IPv6) | ✖ | ✖ | the IPv6 address discovered by the `use` strategy |
+    | ✖ | ✔ | ✖ | the IPv4 address discovered by the `usev4` strategy |
+    | ✖ | ✖ | ✔ | the IPv6 address discovered by the `usev6` strategy |
+    | ✔(IPv4) | ✔ | ✖ | the IPv4 address discovered by the `usev4` strategy (and log another warning if it doesn't match the IPv4 address found by the `use` strategy) |
+    | ✔(IPv6) | ✔ | ✖ | the IPv6 address discovered by the `use` strategy |
+    | ✔(IPv4) | ✖ | ✔ | the IPv4 address discovered by the `use` strategy |
+    | ✔(IPv6) | ✖ | ✔ | the IPv6 address discovered by the `usev6` strategy (and log another warning if it doesn't match the IPv6 address found by the `use` strategy) |
+
+  * To support separate credentials for IPv4 vs. IPv6 updates, convert
+    the `%config` hash of host configs into a list of host configs. A
+    second definition for the same host adds a second entry rather
+    than overwrites the existing entry.
+
+## Alternatives Considered
+
+### Repurpose the existing settings for v4
+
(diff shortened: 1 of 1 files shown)


rhansen:
cc @dkerr64 @abelbeck @cjac

dkerr64:
I'll start with a copy/paste from my original pull request #40...
Adding comprehensive framework for IPv6 support. IPv6 and IPv4 exist side-by-side
The same network interface may have both IPv4 and IPv6. Therefore to support both
we add the following options.

```
usev6 = [ ip, if, cmd, web ]
ipv6 = [ set an IPv6 address (usev6=ip) ]
webv6 = [ url of a web server that will return your IPv6 address (usev6=web)]
webv6-skip = [ string to skip to before looking for IPv6 address]
cmdv6 = [ system command to execute that will return an IPv6 address (usev6=cmd)]
cmdv6-skip = [ string to skip to before looking for IPv6 address ]
```

There is no ifv6. It is assumed that if=interface will be the same for both IPv4
and IPv6

You can mix methods for IPv4 and IPv6 for the same interface. For example
```
use=web
usev6=if
if=eth0
web=http://ipdetect.dnspark.com/
```
Would be a valid method for a system behind an IPv4 NAT connection, IPv6 would be
obtained from the eth0 interface (globaly routable address) and the public IPv4
from external web site.

dkerr64:
Design decision... when usev6=if the IPv6 address selected shall always be a Global Unique Address (GUA) and never a Unique Local Address (ULA) or Link Local (LLA).
However, if a static IPv6 is provided, or is obtained from a cmd script or from a web server, then the IPv6 address used shall be that returned by static/cmd/web.

Side effect of this is that ddclient would not update a "internal" dynamic DNS server (inside campus walls say) with a ULA from a local interface.  If that is required then the IPv6 address shall be obtained by one of the other methods.

dkerr64:
Design decision...  It is not a requirement that we support IPv6 for all Dynamic DNS servers supported by ddclient for IPv4.  We will provide support for a selected few, others may be added later based on demand and whether the Dynamic DNS server supports updating IPv6 addresses (not all do).

dkerr64:
Design decision... All network traffic for IPv6 shall be done through CURL and not through perl's built-in network functions.  This greatly simplifies implementation.  Put another way... I never got it to work and so gave up.

abelbeck:
Question for a perl guru, is there a generally available perl module/method to do the equivalent of `curl -4 ...` and `curl -6 ...` ?  This is a fundamental requirement for adding dual-stack support.

If the answer is yes, but involves a complicated set of requirements that varies over perl versions, then the native perl networking approach may be not as good as simply externally calling `curl`.  Some embedded systems may not have a full set of perl modules but will have `curl`.

IMHO, this question needs to answered early in the process.

dkerr64:
Okay, I just noticed there is a whole design document I should have read first.  Oops, sorry.

I generally agree with everything in there with one exception.  If use=foo is specified and no usev4= nor usev6= then I think it should default to usev4=foo and usev6=no.  In other words the old depreciated settings become equivalent to the new v4 settings and does not infer anything for IPv6.

I think that this would match the vast majority of current users.  The existing IPv6 support is so broken we should require users that need it to use the new implementation.

docs/ipv6-design-doc.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Nice-to-Haves
+
+  * The user should be able to force the update protocol to use IPv4
+    or IPv6.
+  * Unless configured otherwise, ddclient should first attempt to
+    update via IPv6 and fall back to IPv4 if the IPv6 connection
+    fails. This behavior can be added later; for now it is acceptable
+    to keep the current behavior (use IPv6 without IPv4 fallback if
+    there is a AAAA record, use IPv4 if there is no AAAA record).
```
dkerr64, line 81: Why would we not leave this to the underlying network stack in the operating system?

docs/ipv6-design-doc.md
```diff
@@ -0,0 +1,164 @@
...
+  * Full backwards compatibility with existing config files and
+    flags. The trade-off between migration burden and long-term
+    usability should be carefully considered.
```
dkerr64, line 84: This should be mandatory for IPv4.  For IPv6 I'm not sure I care.

docs/ipv6-design-doc.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Configuration changes
+
+  * Add new `usev4` and `usev6` settings that are like the current
+    `use` setting except they only apply to IPv4 and IPv6,
+    respectively.
+      * `usev4` can be set to one of the following values: `ipv4`,
+        `webv4`, `fwv4`, `ifv4`, `cmdv4`, `ciscov4`, `cisco-asav4`
+      * `usev6` can be set to one of the following values: `ipv6`,
+        `webv6`, `fwv6`, `ifv6`, `cmdv6`, `ciscov6`, `cisco-asav6`
```
dkerr64, line 96: I can sign up for ipv6, webv6, ifv6 and cmdv6.  I cannot sign up to implement any of the firewall or gateway methods.

docs/ipv6-design-doc.md
```diff
@@ -0,0 +1,164 @@
...
+    `ifv4` is not set then `if` is used.) Special notes:
+      * The value of `ip` will only serve as the default for `ipv4`
+        (or `ipv6`) if it contains an IPv4 (or IPv6) address.
+      * There is currently an `ipv6` boolean setting. To preserve
+        backward compatibility with existing configs, `ipv6` set to a
+        boolean value is ignored (other than a warning).
+  * Deprecate the `use` setting. Change `use=foo` to control the
+    default for `usev4` and `usev6`:
+      * If neither `usev4` nor `usev6` is set, then `use=foo` is
+        equivalent to `usev4=foov4,usev6=foov6`.
```
dkerr64, line 111: I disagree.  I think the depreciated settings should be equivalent to the new v4 settings.  So `use=foo` is equivalent to `usev4=foov4, usev6=no`. If we don't do this then everyone is going to suddenly be trying to update a IPv6 address at their DDNS server when they probably didn't expect it -- potentially causing errors.

dkerr64:
Also... you should create a separate branch on Github for the IPv6 code to get developed in, and then merge with the master only when you are ready.  I am doing this on my own Github repository right now as a "work-in-progress" and happy to continue with that.  But from time-to-time progress I (or others) make should be merged back into ddclient repository.  But should be in a dedicated branch separate from master until you are ready to pull it in.

dkerr64:
I agree with @SuperSandro2000 as I noted in my comment, depreciated settings should apply for IPv4 only, even if that means some existing IPv6 behavior stops working.

dkerr64:
You do not document it, but the `use` `usev4` and `usev6` may also be sent to the value `no` which means do not obtain an IPv4 (or IPv6) address or update it.

docs/ipv6-design-doc.md
```diff
@@ -105,35 +114,66 @@ Some attempts have been made to add more robust IPv6 support:
...
+| `use` | `usev4` | `usev6` | resulting value |
+| :---: | :---: | :---: | :--- |
+| ✔(IPv4) | ✖ | ✖ | the IPv4 address discovered by the `use` strategy |
+| ✔(IPv6) | ✖ | ✖ | the IPv6 address discovered by the `use` strategy |
+| ✖ | ✔ | ✖ | the IPv4 address discovered by the `usev4` strategy |
+| ✖ | ✖ | ✔ | the IPv6 address discovered by the `usev6` strategy |
+| ✔(IPv4) | ✔ | ✖ | the IPv4 address discovered by the `usev4` strategy (and log a warning if it doesn't match the IPv4 address found by the `use` strategy) |
+| ✔(IPv6) | ✔ | ✖ | the IPv6 address discovered by the `use` strategy |
+| ✔(IPv4) | ✖ | ✔ | the IPv4 address discovered by the `use` strategy |
+| ✔(IPv6) | ✖ | ✔ | the IPv6 address discovered by the `usev6` strategy (and log a warning if it doesn't match the IPv6 address found by the `use` strategy) |
```
dkerr64, line 176: Why bother testing if `use` equals `usev4` or `usev6` and issue a warning?  We should issue the warning whether it matches or not (assuming it is of the same version) so that user is encouraged to remove `use` and replace with the intended replacement.

dkerr64: In fact we should always issue a warning if `use` is set whether or not `usev4` `usev6` are set or match or not.  Setting `use` is always reason for a warning.

rhansen: `use` will always print a warning because it's deprecated. What I mean here is that a second warning would be printed because the IP address discovered by `use` is discarded in favor of the different address discovered by `usev4` or `usev6`.

rhansen:
> You do not document it, but the `use` `usev4` and `usev6` may also be sent to the value `no` which means do not obtain an IPv4 (or IPv6) address or update it.

Wouldn't omitting it be equivalent to `no`?

rhansen:
> Wouldn't omitting it be equivalent to `no`?

Thinking about it more, users will occasionally need to override the global value of `use`, `usev4`, or `usev6`. For example:

```
usev4=ifv4 ifv4=eth0
usev6=ifv6 ifv6=eth0

protocol=dyndns2 \
user=myusername \
password=mypassword \
myhost.example.com

# Google Domains allows a dynamic DNS host to either hold an IPv4 record
# or an IPv6 record but not both. Either `usev4` or `usev6` must be
# disabled to prevent ddclient from throwing an error.
protocol=googledomains \
usev6=disabled \
user=myusername \
password=mypassword \
myhost2.example.com
```

I updated the document to add a new `disabled` strategy. I prefer `disabled` over `no` because it is clearer. (Some users might wonder if `no` is an abbreviation for "number" or something.)

docs/ipv6-design-doc.md
rhansen: The recommended way to do "fallback" is [RFC 8305](https://tools.ietf.org/html/rfc8305). As far as I know there is no Perl module that implements that RFC so we would have to do it ourselves. Or we could use curl, which already implements RFC 8305 (or at least its predecessor, RFC 6555).

dkerr64:
Yes, `disabled` would be a good substitute for `no`.  For compatibility with ddclient-curl we would need to accept both (but can document only `disabled`).

dkerr64:
for `$config{$host}{'wantip'} `I would suggest just replacing it with `wantipv4` with a global find/replace.  No need to keep it around as it is internal variable.

dkerr64:
There is also `$config{$host}{'status'}` and `$config{$host}{'status-ipv6'}` this one is a little more awkward to change to `status-ipv4 `because it is saved in the cache file.  Do cache files persist across execution instances?  If so then we need to accept `status` in a cache read (and change it to `status-ipv4`?)

Or... we should just leave it as `status` and not change to `status-ipv4` but I'm guessing you will not be in favor.

rhansen:
> Yes, `disabled` would be a good substitute for `no`. For compatibility with ddclient-curl we would need to accept both (but can document only `disabled`).

Sounds good to me. I added `no` to the design doc as a deprecated alias of `disabled`.

> for `$config{$host}{'wantip'} `I would suggest just replacing it with `wantipv4` with a global find/replace. No need to keep it around as it is internal variable.

This would break compatibility for users that are currently updating IPv6 addresses. (Compatibility would be restored once the update protocol callbacks are updated to honor `'wantipv6'`.)

> There is also `$config{$host}{'status'}` and `$config{$host}{'status-ipv6'}` this one is a little more awkward to change to `status-ipv4 `because it is saved in the cache file.

`ip` is also cached.

> Do cache files persist across execution instances?

Yes.

> If so then we need to accept `status` in a cache read (and change it to `status-ipv4`?)

No, I don't think we need to. If I understand correctly, after upgrade ddclient would log a warning that says "unrecognized keyword 'status' (ignored)" and maybe "Use of uninitialized value $cache{$host}{'status'}", but those would be one-time warnings (`status` should be replaced with `statusv4` and `statusv6` the next time the cache file is written) and it should continue to work properly. We should test it though.

But I'm not sure that `status` should be split into to `statusv4` and `statusv6`. For dynamic DNS services that can update both types of records with a single HTTP query, a single status probably makes more sense. For dynamic DNS services that require two separate HTTP queries to update the IPv4 and IPv6 records we might need to split `status`, or change `status` to hold multiple values (e.g., `ipv4:good,ipv6:failed`). The same goes for `wtime`, `mtime`, and `atime`. I need to study the code some more to be sure.

dkerr64:
Looking at IO::Socket::IP which is used by HTTP::Tiny it looks like it can force IPv6 (at least for listening sockets).  But I really thing ddclient should use curl for all network traffic.  I don't see a problem with making it required.

dkerr64:
> > for `$config{$host}{'wantip'} `I would suggest just replacing it with `wantipv4` with a global find/replace. No need to keep it around as it is internal variable.
> 
> This would break compatibility for users that are currently updating IPv6 addresses. (Compatibility would be restored once the update protocol callbacks are updated to honor `'wantipv6'`.)

I don't think so.  Handling of depreciated settings should be done outside of the various server/protocol routines -- those should be passed an IPv4 and/or IPv6 addresses in the new variables and not have to handle the intricacies of depreciated wantip.  For example, if the address returned from get_ip is a IPv6 then set it to wantipv6, if a IPv4 set it to wantipv4.

rhansen:
> Looking at IO::Socket::IP which is used by HTTP::Tiny it looks like it can force IPv6 (at least for listening sockets).

True. Unfortunately, I don't see a way to pass IO::Socket::IP options through HTTP::Tiny, so I don't think we can use it.

> But I really think ddclient should use curl for all network traffic. I don't see a problem with making it required.

There is risk to switching to curl. Until we have a good set of unit and integration tests (which I'm working on) I'd like to keep geturl's changes simple. PR #169 doesn't change geturl's current behavior, and I think it will be sufficient for now.

Eventually I would like to switch to curl, for [RFC 8305](https://tools.ietf.org/html/rfc8305) support if nothing else. But I don't want this IPv6 work to be blocked on switching to curl.

> Handling of depreciated settings should be done outside of the various server/protocol routines -- those should be passed an IPv4 and/or IPv6 addresses in the new variables and not have to handle the intricacies of depreciated wantip.

With the proposed design the protocol callbacks do not need any special logic to handle the deprecated `wantip`. Callbacks can either remain as they are right now and continue to use `wantip`, or they can be upgraded to use `wantipv4` and `wantipv6` instead. None of the callbacks should ever use all three. We can delete the old `wantip` after we finish upgrading all of the callbacks.

> For example, if the address returned from get_ip is a IPv6 then set it to wantipv6, if a IPv4 set it to wantipv4.

That would require updating all of the protocol callbacks to support both `wantipv6` and `wantipv4`. I don't want to tie the work of updating all of the callbacks to the IPv6 infrastructure changes. Doing so would result in a huge, complicated PR. By preserving the existing `wantip` value we can upgrade the callbacks over several small PRs.

dkerr64:
The configuration section of the design doc does not document `ifv6-skip`, `webv6-skip` or `cmdv6-skip` or the IPv4 equivalents.  It may be possible to depreciate `if[v6]-skip` and `cmd[v6]-skip` as it could be argued that these are not necessary (output from ip/ifconfig should be left to ddclient to parse, if cmd output needs to be parsed that could be wrapped in a script command) but I can see the need for `web[v6]-skip`.

rhansen:
> The configuration section of the design doc does not document `ifv6-skip`, `webv6-skip` or `cmdv6-skip` or the IPv4 equivalents.

Good eye, thanks. I updated the design doc.

> It may be possible to depreciate `if[v6]-skip` and `cmd[v6]-skip` as it could be argued that these are not necessary (output from ip/ifconfig should be left to ddclient to parse, if cmd output needs to be parsed that could be wrapped in a script command) but I can see the need for `web[v6]-skip`.

I agree with adding `webv4-skip` and `webv6-skip`.

I agree that we shouldn't bother with `ifv4-skip` and `ifv6-skip`—it should be ddclient's responsibility to correctly parse the output of whatever command it uses to get an interface's addresses.

`cmdv4-skip` and `cmdv6-skip` are tougher calls. If we add them, users can use existing utilities without having to write a wrapper script. Also, it'd be easier for users to upgrade their legacy config files. But, as you say, it's easy enough to write wrapper scripts. How about we leave them out for now and add them if someone files a feature request?

dkerr64:
> How about we leave them out for now and add them if someone files a feature request?

Yes.  Keep it simple.

rhansen:
@dkerr64: Thank you for all of the feedback you have provided. What do you think of the design doc right now? Do you agree with it? Do you think it is complete enough?

dkerr64:
@rhansen Yes I think it is good shape.  Other comment/changes that I'd like are outside scope of this document (e.g. provide for option to use curl) and I have or will open issues for those which can be tackled on a case-by-case basis.

dkerr64:
@rhansen There is one problem that is worth discussion.  IPv6 being the nasty beast it is.  A system can (and always does) have multiple IP addresses assigned to an interface.  We have documented that we will ignore ULA and LLA addresses.  But the problem is when we have multiple GUA addresses.

An interface can have a GUA assigned...
1. Statically at the client by the user (ddclient should get this by user assignment in `ipv6` setting)
2. By DHCP from a server (DHCPv6)
3. Auto configuration by router advertisement (RA)
4. Auto configuration by client generating a temporary IPv6

The problem is which one to use.  And worse, we may have no control over it.  For the `ifv6` method we can filter out depreciated and temporary addresses (provided we use `ip` and not `ifconfig` on linux. `ifconfig` on mac however does work).  But should we prioritize a DHCPv6 assigned address over a RA assigned address?  I think we should, but I do not know how to determine that unless you know the DHCPv6 range that the server is allocating from.  You can then look at the lower 64 bits and check whether it is in the DHCPv6 range.  So should we allow the user to pass that in?  By-the-way, I believe that RA is far more common than DHCPv6 on networks behind a firewall/gateway.

And then we have the `webv6` method.  When a device initiates a IPv6 conversation with a web server it is going to pick from one of its temporary IPv6 addresses first.  There can be multiple of these as they are regenerated over time.  This is by design to ensure some privacy... so a website cannot build up a database of known IP addresses associated with, for example, all their customers and make that public.  So IPv6 addresses that a web site sees expire over time so frequent checking will be required.  So for IPv6, the `webv6` method is probably the worst one to use.

And then we get to IPv6 network prefix translation which comes into play when you only have ULAs assigned.  For which the only method available would be the `webv6` method... somewhat like how IPv4 NAT works.

The bottom line is that we need some thorough documentation for users on IPv6, not to discourage its use, but so they are aware of the considerations.

David

abelbeck:
@dkerr64 said:
> here is one problem that is worth discussion. IPv6 being the nasty beast it is. A system can (and always does) have multiple IP addresses assigned to an interface. We have documented that we will ignore ULA and LLA addresses. But the problem is when we have multiple GUA addresses.

True, but for `ddclient` worthy interfaces it is generally simpler.  For example my ISP (via a dhcpv6 client) external interface (`ip -6 -o addr show dev eth0 scope global`) is only assigned a /128 GUA, which is `ddclient` worthy.

If you have internal GUA's with a static prefix (no need for `ddclient`) and you need matching static firewall rules to allow GUA traffic in.

Internal GUA's with a dynamic prefix are messy without dynamic firewall rules to adjust for a changing prefix. While the `ifv6` method can work, most would probably use the `webv6` method, or a custom script as this is an "expert" setup.

The simpler case for internal interfaces and a dynamic prefix is a ULA with NPTv6 (Network Prefix Translation) and the static ULA's can be firewalled statically and the only adjustment for a changing prefix is the NPTv6 mapping.  This requires the `webv6` method with `ddclient`.

TLDR, for the `ifv6` method to be applicable, the first entry outputted from:
```
ip -6 -o addr show dev eth0 scope global
```
will most likely do the trick, filter out ULA's (as they are also `scope global`) to be safe.

dkerr64:
@abelbeck I don't think we can rely upon the order that IPv6 addresses are returned by `ip` or `ifconfig`.  I have tried three systems... MacOS, Ubnutu and RHEL.  I get different results (that is whether a DHCPv6 address or a RA address is returned first) across all three.  On RHEL, `ip` and `ifconfig` return the same two addresses in different order.

abelbeck:
@dkerr64 : I guess with multiple GUA's the one which is least likely to change would be best ... not sure if a simple heuristic could determine that.

Though the "first" GUA `ip` reports should always work as long as the edge firewall forwards the same GUA internally as the AstLinux `dyndns-ipv6-forward` firewall plugin does.

rhansen:
I added some text about how to extract the IPv6 address from the interface. As I mentioned in the document, the good news is we don't have to come up with a perfect solution now—we can implement something that *mostly* works and refine it later.

rhansen:
@wimpunk and @DaveSophoServices: Do you have any thoughts? Or just a thumbs up or down?

rhansen:
I just spent some time looking at freedns.afraid.org, and their v2 API involves two separate update URLs: one for the IPv4 record and one for the IPv6 record. They look like this:

  * IPv4: https://sync.afraid.org/u/token-specific-to-the-A-record-goes-here/?ip=1.2.3.4
  * IPv6: https://v6.sync.afraid.org/u/token-specific-to-the-AAAA-record-goes-here/?ip=dead::beef

(freedns.afraid.org supports alternative update URLs, but I tried them and they're buggy when a single host has both an A record and an AAAA record. Specifically, when updating the IPv4 address it will sometimes convert the AAAA record to an A record, leaving two A records for the same host, one with the old address and one with the new. The same thing happens with IPv6. I contacted them to let them know about this; we'll see what they say.)

I'm not sure how to support this. Maybe we would add freedns-specific `tokenv4` and `tokenv6` settings and substitute them into hard-coded URL templates. But in general I expect other services will have separate v4 vs. v6 update URLs for the same hostname, or separate usernames and passwords for v4 and v6. A general solution that works everywhere would be nice. Thoughts?

dkerr64:
Did you look at how we implemented freedns here... https://github.com/astlinux-project/ddclient-curl/blob/55c623b6fa4686af620871c48bdbcd70048c65f6/ddclient#L3843

As far as I know this has been working well for us.

David

rhansen:
Ah, that's an interesting way to do it. You're using the v1 API which doesn't have the ambiguity problem when you categorize the update URLs by address type.

Unfortunately the v1 API doesn't support specifying the IP address, it just uses the connecting client's IP address. That's why I was investigating the v2 API. Fortunately I got a reply from Josh at freedns.afraid.org and it looks like he's going to fix the address conversion problem.

But the general problem remains: How should we support separate update URLs or separate usernames/passwords for IPv4 vs. IPv6?

dkerr64:
Not sure what you mean... I am absolutely specifying the IP address and specifying `v=2` on the url to ensure I use the v2 api.  It has been working reliably for years.  I just tested it again and was able to update both A and AAAA records.  There is no need to handle separate URLs for IPv4 and IPv6, you can obtain the specific update URL by query which is what I do in the code I linked to.

dkerr64:
Important note... an A or AAAA record must already exist at FreeDNS (should be created through the web site) in order for ddclient to update it.  My ddclient code will not create an A or AAAA record if one does not already exist (I don't think that can be done by api_

dkerr64:
So I have an implementation question.  Right now we have a function `get_ip` that could return a IPv4 or IPv6... there is nothing to prevent any method from returning an IPv6 from, e.g the `web`, `ip` or `cmd` method and either an IPv4 or IPv6 is considered valid.

My current implementation (over in ddclient-curl) adds a new function `get_ipv6` that is guaranteed to return a IPv6 address (or undef) and modifies `get_ip` to always get a IPv4.  This potentially breaks backward compatibility for settings that we intend to depreciate.  So the easiest thing to do, I think, is to leave `get_ip` as it is and add a new `get_ipv4` that is guaranteed to return an IPv4 address (or undef).  We would therefore have three functions that we would call get an IP address based on settings `use`, `usev4` and `usev6`.  The `get_ipv4` and `get_ipv6` would always take precedence over `get_ip`.

Comments?
David

rhansen:
Yup, that's exactly what I was envisioning.

dkerr64:
As I added all the new settings it raised some questions.  See here in the pull request for that... https://github.com/ddclient/ddclient/blob/8d0d168a5603592eaee9dfbb2e03ff941003b0ce/ddclient#L860

Specifically whether we should introduce `fwv4-skip` and `fwv6-skip` settings to mirror the existing `fw-skip`.  And also whether we need to do the same for `fw-banlocal`.

Right now I have put the skip settings in (but commented out) but not the banlocal -- partly because for the banlocal we have sort-of already declared that we will accept IPv6 GUA addresses only.  But is that reasonable assertion for when we use the firewall method?  I have not looked at the existing code for firewall method.  And I do not have access to any cisco firewalls.  And I don't even know if it is reasonable to want to acquire an IPv6 address from a firewall like ddclient has done in the past for IPv4.

Unless someone has access to the necessary equipment and is willing to research and design this, I propose removing the firewall method for IPv6.

David

rhansen:
> But the general problem remains: How should we support separate update URLs or separate usernames/passwords for IPv4 vs. IPv6?

I added some text to the design document to address this. tl;dr: Allow users to define the same host multiple times in their config file. This would allow them to write their configs like this:

```
usev4=if ifv4=eth0 \
protocol=foo \
login=my-ipv4-username \
password=my-ipv4-username \
host.example.com

usev6=if ifv6=eth0 \
protocol=foo \
login=my-ipv6-username \
password=my-ipv6-username \
host.example.com
```

> Specifically whether we should introduce `fwv4-skip` and `fwv6-skip` settings to mirror the existing `fw-skip`.

I was assuming we would. But we can add those later; I suspect not many people use them so they're low priority.

> And also whether we need to do the same for `fw-banlocal`.

`fw-banlocal` is problematic:
  * There's not much point to it. Regardless of whether it is enabled, the end result is a DNS record that is not being updated to a useful value.
  * It's a misnomer: `fw-banlocal` doesn't just filter out local IP addresses from `use=fw`, it also filters them out of all other address sources except `use=ip`.
  * It doesn't filter out local IPv6 addresses.
  * The resulting warning ("unable to determine IP address") is misleading.

I think we should deprecate `fw-banlocal` and not port it over to the new `usev4` and `usev6` logic.

We might want to log a warning if we are updating to a non-global address—perhaps with an option to silence those warnings—but that can be decided later.

> Right now I have put the skip settings in (but commented out) but not the banlocal -- partly because for the banlocal we have sort-of already declared that we will accept IPv6 GUA addresses only.

We should ignore temporary or non-GUA addresses only for `ifv6`. For all other mechanisms (`cmdv6`, `webv6`, `fwv6`, etc.) we should accept any IPv6 address regardless of its type.

rhansen:
I'm going to merge this tomorrow unless there are any objections.

PR #166, opened 2020-06-10
Bump minimum required Perl to v5.10.1

rhansen:
This allows us to use the `//` and `//=` operators.

v5.10.1 was chosen because that is the oldest version of Perl among all currently supported releases of Ubuntu, CentOS, RHEL, Fedora, and Debian.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -11,7 +11,7 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
 
 ### Compatibility changes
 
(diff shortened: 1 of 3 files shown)


SuperSandro2000:
I think you wanna rebase this on master.

rhansen:
It's currently rebased on top of PR #158 so that this PR can update the ChangeLog to note the new Perl dependency. Once PR #158 is merged I'll rebase this onto master.

PR #168, opened 2020-06-11
Change geturl to take a hash of parameters

rhansen:
This makes the call sites more readable, and it will be easier to extend in the future (to add an option to force IPv4 or IPv6, for example).


rhansen:
I'm going to combine this with PR #169.

PR #169, opened 2020-06-11
New geturl param to force IPv4 or IPv6

rhansen:
This will be used by the upcoming `webv4` and `webv6` options to ensure that the checkip service returns the desired type of IP address.

Addresses #172 

cc @dkerr64 @abelbeck

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -1521,7 +1521,12 @@ sub test_possible_ip {
 sub test_geturl {
     my $url = shift;
 
-    my $reply = geturl(opt('proxy'), $url, opt('login'), opt('password'));
+    my $reply = geturl({
+        proxy => opt('proxy'),
+        url => $url,
+        login => opt('login'),
+        password => opt('password'),
+    });
     print "URL $url\n";
     print defined($reply) ? $reply : "<undefined>\n";
     exit;
@@ -1995,13 +2000,14 @@ EOM
 ## geturl
 ######################################################################
 sub geturl {
-    my $proxy    = shift || '';
-    my $url      = shift || '';
-    my $login    = shift || '';
-    my $password = shift || '';
-    my $headers  = shift || '';
-    my $method   = shift || 'GET';
-    my $data     = shift || '';
+    my ($params) = @_;
+    my $proxy = $params->{proxy} // '';
+    my $url = $params->{url} // '';
+    my $login = $params->{login} // '';
+    my $password = $params->{password} // '';
+    my $headers = $params->{headers} // '';
+    my $method = $params->{method} // 'GET';
+    my $data = $params->{data} // '';
     my ($peer, $server, $port, $default_port, $use_ssl);
     my ($sd, $request, $reply);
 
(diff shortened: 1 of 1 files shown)


ddclient
```diff
@@ -1988,128 +1993,133 @@ EOM
...
 ######################################################################
 sub geturl {
-    my $proxy    = shift || '';
-    my $url      = shift || '';
-    my $login    = shift || '';
-    my $password = shift || '';
-    my $headers  = shift || '';
-    my $method   = shift || 'GET';
-    my $data     = shift || '';
+    my ($p) = @_;
```
SuperSandro2000, line 1996: I am not a big fan of the single p variable. Maybe we find a good three letter one?

rhansen: You haven't written much Go code, have you? :stuck_out_tongue_winking_eye: 

I decided to keep the original variables.

ddclient
```diff
@@ -2068,56 +2075,59 @@ sub geturl {
...
-            MultiHomed => 1,
-            Timeout    => opt('timeout'),
-        );
-        defined $sd or warning("cannot connect to %s:%s socket: %s %s", $peer, $port, $@, IO::Socket::SSL::errstr());
-    } elsif ($globals{'ipv6'}) {
+    my $socket_class = IO::Socket::INET;
+    if ($use_ssl) {
+        # IO::Socket::SSL will load IPv6 support if available on the system.
+        load_ssl_support;
+        $socket_class = IO:Socket::SSL;
```
SuperSandro2000, line 2082:
```suggestion
        $socket_class = IO::Socket::SSL;
```

rhansen: Whoops, thank you.

SuperSandro2000:
Also you did not test this very well:

```
Bareword "IO::Socket::INET" not allowed while "strict subs" in use at ./ddclient line 2078.
Bareword "IO::Socket::SSL" not allowed while "strict subs" in use at ./ddclient line 2082.
Bareword "IO::Socket::INET6" not allowed while "strict subs" in use at ./ddclient line 2085.
Execution of ./ddclient aborted due to compilation errors.
```

rhansen:
> Also you did not test this very well:

No, I didn't. If we think that PR #161 will be merged soon I'll write some unit tests for this.

SuperSandro2000:
You should test it anyway cause right now this breaks ddclient and I can't merge it.

rhansen:
I did run some manual tests with all combinations of with/without SSL and unforced/force-v4/force-v6 and they all worked.

I'm also writing unit tests so that we can be comfortably confident. I'm going to change this to draft until the tests have been written.

dkerr64:
I've also tested with forced/unforced and it works.  It is notable that you do get different results between unforced and the forced for the web method.  Some websites that report your IP do it based on whether you connect over IPv4 or IPv6.  With the unforced method you have no control over this... it is left up to the underlying network stack in the operating system and it could decide to use IPv4 or IPv6 on a whim (well, actually I think the rules say to bias towards IPv6 unless IPv4 is going to be faster).  But it does mean that the results of a web method to the same site could change from time-to-time.

The bottom line being that the quicker we can depreciate the old settings the better.

David

rhansen:
Testing this has turned out to be more obnoxious than I had hoped, and I don't want to hold up this PR while I work out the kinks. I've manually tested all of the cases and believe the behavior to be correct.

PR #171, opened 2020-06-12
Prefer `ip` command from iproute2 over `ifconfig`

### LINKED ISSUE #93: use iproute2 by default, with net-tools as fallback
Hi,
as the title suggest, I think it would be a good time now to switch away from net-tools. this has been deprecated for years in favor of iproute2, and is only still working nowadays because the kernel did not break it... Yet.

I understand you can't change it, as in permanently get rid of this, but perhaps ddclient could start using either net-tools or iproute2 where possible, and not exclusively rely on net-tools.

rhansen:
On Linux systems, `ifconfig` is long deprecated in favor of the `ip` command from iproute2. Some systems don't have iproute2 (BSDs in particular), so ddclient will still attempt `ifconfig` if `ip` is missing.

Also: Don't hide STDERR because error messages are important for troubleshooting problems. To avoid STDERR noise on systems without the `ip` command, the command's existence is checked before it is run.

Notes:
  * The fetched addresses could be limited to IPv4 or IPv6 depending on `opt('ipv6')`, and non-global addresses could be filtered out, but any filtering risks breaking a nontrivial number of existing configurations.
  * This change runs the risk of breaking existing configs that set `if-skip`. Due to the deprecation of `ifconfig`, and the belief that only a negligible number of users set `if-skip`, the benefits of this change are believed to outweigh the config migration burden imposed on users.

Fixes #93.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -754,7 +754,7 @@ my @opt = (
     [ "ip",          "=s", "-ip address           : set the IP address to 'address'" ],
     "",
     [ "if",          "=s", "-if interface         : obtain IP address from 'interface'" ],
-    [ "if-skip",     "=s", "-if-skip pattern      : skip any IP addresses before 'pattern' in the output of ifconfig {if}" ],
+    [ "if-skip",     "=s", "-if-skip pattern      : skip any IP addresses before 'pattern' in the output of 'ip address show dev {if}' (or 'ifconfig {if}')" ],
     "",
     [ "web",         "=s", "-web provider|url     : obtain IP address from provider's IP checking page" ],
     [ "web-skip",    "=s", "-web-skip pattern     : skip any IP addresses before 'pattern' on the web provider|url" ],
@@ -1477,8 +1477,15 @@ sub test_possible_ip {
 
     {
         local $opt{'use'} = 'if';
-        foreach my $if (grep {/^[a-zA-Z]/} `ifconfig -a`) {
-            $if =~ s/:?\s.*//is;
+        # Note: The `ip` command adds a `@eth0` suffix to the names of VLAN
+        # interfaces. That `@eth0` suffix is NOT part of the interface name.
+        my @ifs = map({ /^[^\s:]*:\s*([^\s:@]+)/ ? $1 : () }
+                      `command -v ip >/dev/null && ip -o link show`);
+        @ifs = map({ /^([a-zA-Z].*?)(?::?\s.*)?$/ ? $1 : () }
+                   `command -v ifconfig >/dev/null && ifconfig -a`) if $? || !@ifs;
+        @ifs = () if $?;
+        warning("failed to get list of interfaces") if !@ifs;
+        foreach my $if (@ifs) {
             local $opt{'if'} = $if;
             printf "use=if, if=%s address is %s\n", opt('if'), define(get_ip('if'), 'NOT FOUND');
         }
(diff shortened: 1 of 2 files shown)


rhansen:
cc @dkerr64 @xogium

ddclient
```diff
@@ -1475,8 +1475,11 @@ sub test_possible_ip {
 
     {
         local $opt{'use'} = 'if';
-        foreach my $if (grep {/^[a-zA-Z]/} `ifconfig -a`) {
-            $if =~ s/:?\s.*//is;
+        my @ifs = map({ /^[^\s:]*:\s*([^\s:]+):/ ? $1 : () } `ip -o link show 2>/dev/null`);
```
dkerr64, line 1478: This fails for VLANS where ip returns the interface with a @ sign between the name and the base interface that the VLAN is attached to.  Testing your code I get the following output...

```
eth0
eth1
eth2
eth3
eth1.100@eth1
eth1.200@eth1
eth1.300@eth1
```

So for each of the VLANS the regex needs to terminate at the @ symbol (or the colon).  I'm not very good with regex so don't have a quick fix for you.

ddclient
```diff
@@ -1475,8 +1475,11 @@ sub test_possible_ip {
...
+        @ifs = map({ /^([a-zA-Z].*?)(?::?\s.*)?$/ ? $1 : () } `ifconfig -a`) if $? || !@ifs;
```
dkerr64, line 1479: you need to add `2>/dev/null` to suppress any error message from ifconfig.  I deliberately tested for failure here and got a double warning, one from ifconfig stderr and the one in your code.

dkerr64:
I'm surprised you pushed this without addressing the VLAN bug.

ddclient
rhansen: Wow, that's poor design. Why does it show `@eth1` there when that's not actually part of the interface name?  :angry:

Fixed, thanks for bringing it up.

ddclient
rhansen: I intentionally didn't redirect STDERR to /dev/null. In fact, I just changed the `if` command to remove the redirection.

STDERR is important for troubleshooting. Redirecting it to /dev/null is a code smell—there's usually a way to suppress expected errors without also suppressing unexpected errors.

dkerr64: I think it is better the way you did it, with `command -v`
Thanks
David

ddclient
```diff
@@ -1475,8 +1475,15 @@ sub test_possible_ip {
 
     {
         local $opt{'use'} = 'if';
-        foreach my $if (grep {/^[a-zA-Z]/} `ifconfig -a`) {
-            $if =~ s/:?\s.*//is;
+        # Note: The `ip` command adds a `@eth0` suffix to the names of VLAN
+        # interfaces. That `@eth0` suffix is NOT part of the interface name.
+        my @ifs = map({ /^[^\s:]*:\s*([^\s:@]+)/ ? $1 : () }
+                      `command -v ip >/dev/null && ip -o link show`);
```
SuperSandro2000, line 1481:
```
command -v ip >/dev/null && ip -o link show
Dump was interrupted and may be inconsistent.
....
a gazillion networks
....
```

I got this during testing. Not sure what it means and I couldn't find something on the net. Do we need to worry about this?

rhansen: I would like to understand what happened in case it's a new bug. Were you using `-debug` and/or `-verbose`?  Note that `-debug` causes it to continue to attempt the update.

SuperSandro2000: I just put the command into my shell and I think I know what happend: I just have to many rules that change sometimes.

PR #173, opened 2020-06-13
Don't replace LF with CRLF in body of HTTP request

rhansen:
Now the Content-Length header matches the actual length of the body.

Also, don't replace an LF with CRLF if it is already preceeded by CR.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -2004,7 +2004,7 @@ sub geturl {
     my $method   = shift || 'GET';
     my $data     = shift || '';
     my ($peer, $server, $port, $default_port, $use_ssl);
-    my ($sd, $rq, $request, $reply);
+    my ($sd, $request, $reply);
 
     ## canonify proxy and url
     my $force_ssl;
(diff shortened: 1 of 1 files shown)


ddclient
```diff
@@ -2063,10 +2063,11 @@ sub geturl {
     $headers .= "\n" if $headers ne '' && substr($headers, -1) ne "\n";
     $request .= $headers;
     $request .= "\n";
+    # RFC 7230 says that all lines before the body must end with <cr><lf>.
+    (my $rq = $request) =~ s/(?<!\r)\n/\r\n/g;
```
SuperSandro2000, line 2067: Why do we need ``$rq`` exactly? It is only used at one other place and I think it is redundant.

rhansen: `$rq` is what is actually sent to the server. `$request` is only used for log messages.

rhansen: Mostly I didn't want to change too much in one commit, though including `\r` in syslog, stdout, and/or emailed logs also seems risky.

PR #174, opened 2020-06-13
Fix global fallback when host-specific setting is undefined

rhansen:
This commit only changes behavior if the hostname is a falsy string (`''` or `'0'`) and there is a variable that:
  * was not set globally before the host definition,
  * was not set in the host definition,
  * was set globally after the host definition, and
  * is relevant to the host.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -1708,7 +1708,7 @@ sub minimum {
 sub opt {
     my $v = shift;
     my $h = shift;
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
So it no longer treats falsy strings as hostnames?

rhansen:
It no longer pays attention to the truthiness of the hostname, only the definedness.

PR #175, opened 2020-06-13
Improve parsing of config file assignments

rhansen:
* Ignore empty `key=value` strings in variable assignments. Now the following lines are supported with the expected meaning:
      * `a=1 , b=2`
      * `a=1,,b=2`
  * Improve the warning message when there is an unterminated quote.
  * Add a warning if the line ends with a backslash.

Also add unit tests.

This is marked as draft because it is based on PR #161. It will be rebased onto master and marked as ready to review after PR #161 is merged.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -30,8 +30,17 @@ m4_foreach_w([_m], [
   ], [AX_PROG_PERL_MODULES([_m], [],
          [AC_MSG_ERROR([missing required Perl module _m])])])
 
-# Perl modules required for tests. Only prints a warning if not
-# installed.
+# Perl modules required for tests. If these modules are not installed
+# then some tests will fail. Only prints a warning if not installed.
+m4_foreach_w([_m], [
+    Data::Dumper
+    Test::More
+  ], [AX_PROG_PERL_MODULES([_m], [],
+         [AC_MSG_WARN([some tests will fail due to missing module _m])])])
+
+# Optional Perl modules for tests. If these modules are not installed
+# then some tests will be skipped, but no tests should fail. Only
+# prints a warning if not installed.
 m4_foreach_w([_m], [
     Test::Warnings
   ], [AX_PROG_PERL_MODULES([_m], [],
--- a/t/parse_assignments.pl
+++ b/t/parse_assignments.pl
@@ -1,0 +1,55 @@
+use Test::More;
+use Data::Dumper;
+
+SKIP: { eval { require Test::Warnings; } or skip($@, 1); }
+eval { require 'ddclient'; } or BAIL_OUT($@);
+
+$Data::Dumper::Sortkeys = 1;
+
+sub tc {
+    return {
+        name => shift,
+        input => shift,
+        want_vars => shift,
+        want_rest => shift,
+    };
+}
+
+my @test_cases = (
+    tc('no assignments',             "",             {},                     ""),
+    tc('one assignment',             "a=1",          { a => '1' },           ""),
+    tc('empty value',                "a=",           { a => '' },            ""),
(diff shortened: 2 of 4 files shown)


dkerr64:
Not sure if this is the right place to comment, but years ago I discovered that the config file format is kind of weird and not at all robust when it comes to including multiple services.  Take this example...
```
ssl=yes
protocol=cloudflare
server=api.cloudflare.com/client/v4
zone=example.com
login=user@example.com
password=secret
use=if
if=eth0
myserver.example.com

protocol=freedns
use=if
if=eth0
login=user
password=secret
myserver.example.com
```
Now I might have expected this to work, but it fails because the FreeDNS protocol attempts to access `api.cloudflare.com/client/v4` despite ddclient documenting that the default is `freedns.afraid.org`.  So to make the above work you have to add the line.
```
server=freedns.afraid.org
```
and if you don't want ssl for FreeDNS then would have to explicitly set that to no.

I suspect that when ddclient was originally written multiple services were not anticipated.  If possible I think we should implement support for a more standard conf/ini file format along the lines of...
```
[defaults]
use=if
if=eth0

[service]
protocol=cloudflare
server=api.cloudflare.com/client/v4
zone=example.com
login=user@example.com
password=secret
ssl=yes
domain=myserver.example.com

[service]
protocol=freedns
login=user
password=secret
domain=myserver.example.com
```
Each `[service]` would inherit from `[defaults]` but would never inherit from another service definition.

rhansen:
> Now I might have expected this to work, but it fails because the FreeDNS protocol attempts to access `api.cloudflare.com/client/v4` despite ddclient documenting that the default is `freedns.afraid.org`. So to make the above work you have to add the line.
> 
> ```
> server=freedns.afraid.org
> ```

Yup, that's known behavior. As far as I can tell, that behavior grew organically over time and is that way for backwards-compatibility reasons.

A more robust way to write your config is to prefix your hostname declaration with the options that should only apply to that host:

```
# globals that apply to all accounts (note the lack of backslashes)
ssl=yes
use=if
if=eth0

# my cloudflare account (note the backslashes):
protocol=cloudflare \
server=api.cloudflare.com/client/v4 \
zone=example.com \
login=user@example.com \
password=secret \
myserver.example.com

# my freedns account (note the backslashes):
protocol=freedns \
login=user \
password=secret \
myserver.example.com
```

> If possible I think we should implement support for a more standard conf/ini file format along the lines of...

Sounds good to me, if we can retain backwards compatibility. Would you mind opening a new issue for this?

dkerr64:
Issue #189 opened to support a more standard config file format.

t/parse_assignments.pl
```diff
@@ -0,0 +1,55 @@
+use Test::More;
+use Data::Dumper;
```
SuperSandro2000, line 2: Probably all distributions ship this core module but we might wanna add it anyway to the test requirements just to make sure none of these drop them in the future.

rhansen: Done.

configure.ac
```diff
@@ -30,8 +30,17 @@ m4_foreach_w([_m], [
...
 
-# Perl modules required for tests. Only prints a warning if not
-# installed.
+# Perl modules required for tests. If these modules are not installed
+# then some tests will fail. Only prints a warning if not installed.
+m4_foreach_w([_m], [
+    Data::Dumper
+    Test::More
+  ], [AX_PROG_PERL_MODULES([_m], [],
+         [AC_MSG_WARN([some tests will fail due to missing module _m])])])
```
SuperSandro2000, line 39: We missed the missing find in the first PR where we added fedora. Maybe an error would be better to not puzzle why the tests failed but straight up know that we miss something.

rhansen: I'll do that in a separate PR.

Issue #177, opened 2020-06-14
Does ddclient really need Data::Validate::IP


dkerr64:
Is it really necessary to require module Data::Validate::IP.  This is not installed by default and I think that ddclient should be as lightweight as possible.  Requiring this is going to cause ddclient to fail on many systems, especially lightweight embedded systems.  I ran into this myself on Ubuntu 20.04.

The only thing this is used for is to check for a valid IPv6.  There is already a check_value() function in ddclient and it could easily be enhanced to validate a IPv6 address, as implemented here... https://github.com/astlinux-project/ddclient-curl/blob/55c623b6fa4686af620871c48bdbcd70048c65f6/ddclient#L1964

If all we need this is to validate an IP address then we should do it ourselves.

David.

rhansen:
I was wondering the same thing. There is no official Data::Validate::IP RPM for CentOS 6 or 7 and their RHEL counterparts, so requiring that dependency immediately rules out those platforms without extra effort on the part of the packager.

dkerr64:
I vote for removing this dependency ASAP.

rhansen:
Sounds good to me. Mind putting together a pull request?

PR #178, opened 2020-06-15
Remove dependency on Data::Validate::IP 

dkerr64:
The Data::Validate::IP module is not available by default on many systems, and not available at all on lightweight embedded systems.  As all we need it for is to validate IPv6 addresses it is better that we do this in our own function and remove this dependency.  Resolves https://github.com/ddclient/ddclient/issues/177

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -24,7 +24,6 @@ use warnings;
 use Getopt::Long;
 use Sys::Hostname;
 use IO::Socket;
-use Data::Validate::IP;
 
 my $version  = "3.9.1";
 my $programd = $0;
@@ -2304,6 +2303,15 @@ sub get_ip {
 
     debug("get_ip: using %s, %s reports %s", $use, $arg, define($ip, "<undefined>"));
     return $ip;
+}
+
+######################################################################
+## is_ipv6() validates if string is valid IPv6 address
+######################################################################
+sub is_ipv6 {
+    my ($value) = @_;
+    # This little gem from http://home.deds.nl/~aeron/regex/
+    return $value =~ /^(((?=.*(::))(?!.*\3.+\3))\3?|([\dA-F]{1,4}(\3|:\b|$)|\2))(?4){5}((?4){2}|(((2[0-4]|1\d|[1-9])?\d|25[0-5])\.?\b){4})\z/ai;
 }
 
 ######################################################################
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -8,6 +8,7 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
 ### New features
 
   * Added support for OVH DynHost.
+  * Removed dependency on Data::Validate::IP.
 
 ### Compatibility changes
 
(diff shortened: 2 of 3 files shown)


rhansen:
Please update README.md and add a note in ChangeLog.md.

rhansen:
Actually I can do that.

ddclient
```diff
@@ -2306,6 +2305,16 @@ sub get_ip {
...
 }
 
+######################################################################
+## is_ipv6() validates if string is valid IPv6 address
+######################################################################
+sub is_ipv6 {
+    my $value = shift;
+    # This little gem from http://home.deds.nl/~aeron/regex/
+    return undef if $value !~ /^(((?=.*(::))(?!.*\3.+\3))\3?|([\dA-F]{1,4}(\3|:\b|$)|\2))(?4){5}((?4){2}|(((2[0-4]|1\d|[1-9])?\d|25[0-5])\.?\b){4})\z/ai;
+    return $value
```
rhansen, line 2315: Why not just return the result of the bind? Like this:

```suggestion
    return $value !~ /^(((?=.*(::))(?!.*\3.+\3))\3?|([\dA-F]{1,4}(\3|:\b|$)|\2))(?4){5}((?4){2}|(((2[0-4]|1\d|[1-9])?\d|25[0-5])\.?\b){4})\z/ai;
```

dkerr64: It depends if we want to replicate the behavior of `is_ipv6` in Data::Validate:IP or do our own thing.  If replicating then the function should return `undef` or the valid IPv6 as a string, which is what I implemented.  The actual code in `is_ipv6` does this...
```
## no critic (RegularExpressions::ProhibitCaptureWithoutTest)
$value =~ /(.+)/;
return $1;
```
Which I don't really understand why that is necessary (I am not a Perl expert).  Do you know?

However if we want to return a true/false (which is perfectly fine with me, and I think more correct) then we change to your suggestion... though `!~` needs to change to` =~` 

Let me know what you prefer (I vote to change)

David

rhansen: Let's change it.

> though !~ needs to change to =~

Oops, yup.

rhansen: Actually, I noticed that there's a capture group around the whole IPv6 address, so how about this:
```suggestion
    $value =~ /^(((?=.*(::))(?!.*\3.+\3))\3?|([\dA-F]{1,4}(\3|:\b|$)|\2))(?4){5}((?4){2}|(((2[0-4]|1\d|[1-9])?\d|25[0-5])\.?\b){4})\z/ai;
    return $1;
```
It's one line longer, but it has the same behavior as `is_ipv6`.

rhansen: Ugh, nevermind. The RE is awfully complicated and I don't think I adjusted it properly (that `(?4)` might need to be `(?5)`, not sure). Probably best to just return bool.

rhansen: I went ahead and made the change.

dkerr64:
Thank you Richard.

rhansen:
Thank you very much for putting this together, and for your work on IPv6 stuff in general. Your contribution is appreciated!

PR #184, opened 2020-06-17
Redo freedns.afraid.org protocol to fix several bugs

### LINKED ISSUE #180: freedns.afraid.org implementation is buggy
ddclient's implementation of the `freedns` protocol has some bugs and limitations:

  * The desired IP address is never passed to freedns.afraid.org—it assumes the desired IP address matches the client IP address as seen by freedns.afraid.org.
  * IPv6 is not supported.
  * The [v1 update protocol](https://freedns.afraid.org/api/) is used, which requires two requests: one to get the update URL for the desired host and another to actually perform the update. With [API v2](https://freedns.afraid.org/dynamic/v2/), a few different URLs are supported, and we can use whichever one we find most convenient. All of them only require a single request. There is a caveat; see below for details.
  * Failures to update the host are not always recorded in the host's `status`.
  * If one host update fails it might skip other hosts.
  * Non-OK codes from the web server are ignored.
  * Headers are not stripped before parsing the response.

The [implementation in ddclient-curl](https://github.com/astlinux-project/ddclient-curl/blob/55c623b6fa4686af620871c48bdbcd70048c65f6/ddclient#L3844) looks to be cleaner and more robust than ddclient's implementation. It also uses the v1 protocol. (Note: The `&v=2` in the update URL only turns on IPv6 support; it does not signify API v2.)

API v2 caveat: If there is an A record and a AAAA record for the same host and an IPv4 address is supplied via one of the three non-token URLs (see below), freedns.afraid.org might convert the AAAA record to to an A record and update that instead of the pre-existing A record. This leaves two A records for the same host, and they might have different addresses. Similarly, sending an IPv6 address might cause the A record to switch to an AAAA record, leaving two AAAA records. I reported this to Josh at afraid.org and he proposed a change to make it work as expected. Until that change is made, it is probably best to continue to use the v1 API.

API v2 supports these four URLs:
  * `https://sync.afraid.org/u/<record-token>/?ip=<address>`
  * `https://sync.afraid.org/u/?u=<user>&p=<password>&h=<hostname>&ip=<address>`
  * `https://<user>:<password>@sync.afraid.org/u/?h=<hostname>&ip=<address>` (basic auth)
  * `https://<user>:<password>@sync.afraid.org/nic/update/?hostname=<hostname>&myip=<address>` (basic auth)

sync.afraid.org only has an A record. To reach the update service via IPv6 you have to use v6.sync.afraid.org. Both hosts can update both record types (A or AAAA); the only difference is the auto-detected IP address if the `&ip=` query parameter is left out.

cc @dkerr64

rhansen:
* Support IPv6 addresses.
  * Support updating addresses that aren't the client's own addresses.
  * Set status to 'failed' if the update fails for any reason.
  * Don't skip hosts if a previous update failed.
  * Check for a non-OK code from the update server.
  * Strip headers before processing responses.

This still uses API v1 because API v2 currently has some limitations; see #180 for details.

Fixes #180

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -14,8 +14,15 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
 
         ./autogen && ./configure && make && make check && make install
 
+  * The `freedns` protocol (for https://freedns.afraid.org) now supports IPv6
+    addresses.
+
 ### Bug fixes
 
+  * Minor `freedns` protocol fixes, including:
+      * You can now update an address that differs from the system's own.
+      * If multiple hosts are defined and one fails, ddclient will no longer
+        skip the remaining hosts.
   * Fixed a regression introduced in v3.9.0 that caused
     `use=ip,ip=<ipv4-address>` to fail.
   * "true" is now accepted as a boolean value.
--- a/ddclient.template
+++ b/ddclient.template
@@ -3911,62 +3911,96 @@ EoEXAMPLE
 ######################################################################
 ## nic_freedns_update
 ##
-## written by John Haney
+## API v1 documented at http://freedns.afraid.org/api/
 ##
-## based on http://freedns.afraid.org/api/
-## needs this url to update:
-## http://freedns.afraid.org/api/?action=getdyndns&sha=<sha1sum of login|password>
-## This returns a list of host|currentIP|updateURL lines.
-## Pick the line that matches myhost, and fetch the URL.
-## word 'Updated' for success, 'fail' for failure.
+## An update requires two steps. The first is to get a list of records from:
+##   http://freedns.afraid.org/api/?action=getdyndns&v=2&sha=<sha1sum of login|password>
+## The returned list looks like:
 ##
+##   hostname1.example.com|1.2.3.4|http://example/update/url1
+##   hostname1.example.com|dead::beef|http://example/update/url2
+##   hostname2.example.com|5.6.7.8|http://example/update/url3
+##   hostname2.example.com|9.10.11.12|http://example/update/url4
+##   hostname3.example.com|cafe::f00d|http://example/update/url5
+##
+## The record's columns are separated by '|'. The first is the hostname, the second is the current
+## address, and the third is the record-specific update URL. There can be multiple records for the
+## same host, and they can even have the same address type. Any record can be updated to hold
+## either type of address (e.g., if given an IPv6 address the record will automatically become an
+## AAAA record).
+##
+## The second step is to visit the appropriate record's update URL with
+## ?address=<ipv4-or-ipv6-address> appended. "Updated" in the result means success, "fail" means
+## failure.
 ######################################################################
 sub nic_freedns_update {
-
-
     debug("\nnic_freedns_update -------------------");
-
-    ## First get the list of updatable hosts
-    my $url;
-    $url = "http://$config{$_[0]}{'server'}/api/?action=getdyndns&sha=" . &sha1_hex("$config{$_[0]}{'login'}|$config{$_[0]}{'password'}");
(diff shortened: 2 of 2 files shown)


ddclient
```diff
@@ -3827,70 +3827,97 @@ EoEXAMPLE
 ######################################################################
 ## nic_freedns_update
 ##
-## written by John Haney
```
SuperSandro2000, line 3830: I am not sure if we want to keep those or not. Generally I am not a big fan of them cause most sections are copied from other providers.

SuperSandro2000:
I think this should close #180, right?

ddclient
rhansen: I'm not a fan of them either for that reason and because they become less and less true over time. Credit should go in the ChangeLog and commit messages.

rhansen:
> I think this should close #180, right?

Done.

SuperSandro2000:
```
=== cache ====
DEBUG:    get_ip: using cmd, /home/hotpi/src/ddclient/echo reports 10.0.0.1
INFO:     forcing updating myhost.freedns.afraid.org because no cached entry exists.
DEBUG:
DEBUG:    nic_freedns_update -------------------
DEBUG:    proxy    =
DEBUG:    protocol = https
DEBUG:    server   = freedns.afraid.org
DEBUG:    url      = api/?action=getdyndns&v=2&sha=cc020f37f9b10139c288c19d976402110e4d419f
CONNECT:  freedns.afraid.org
CONNECTED:  using SSL
SENDING:  GET /api/?action=getdyndns&v=2&sha=cc020f37f9b10139c288c19d976402110e4d419f HTTP/1.0
SENDING:  Host: freedns.afraid.org
SENDING:  User-Agent: ddclient/3.9.1
SENDING:  Connection: close
SENDING:
RECEIVE:  HTTP/1.1 200 OK
RECEIVE:  Server: nginx
RECEIVE:  Date: Thu, 18 Jun 2020 10:07:27 GMT
RECEIVE:  Content-Type: text/html; charset=UTF-8
RECEIVE:  Connection: close
RECEIVE:  Vary: Accept-Encoding
RECEIVE:  X-Cache: MISS
RECEIVE:
RECEIVE:  ERROR: Could not authenticate.
INFO:     myhost.freedns.afraid.org: setting IP address to 10.0.0.1
FAILED:   updating myhost.freedns.afraid.org: host record does not exist
```
This should now write that the record does not exist but that the login failed.
Because without verbose it looks like this and is not informing the end user correctly
```
DEBUG:    get_ip: using cmd, /home/hotpi/src/ddclient/echo reports 10.0.0.1
DEBUG:
DEBUG:    nic_freedns_update -------------------
DEBUG:    proxy    =
DEBUG:    protocol = https
DEBUG:    server   = freedns.afraid.org
DEBUG:    url      = api/?action=getdyndns&v=2&sha=cc020f37f9b10139c288c19d976402110e4d419f
FAILED:   updating myhost.freedns.afraid.org: host record does not exist
```

rhansen:
> This should now write that the record does not exist but that the login failed. Because without verbose it looks like this and is not informing the end user correctly

Good point, fixed.

SuperSandro2000:
```
FAILED:   updating myhost.freedns.afraid.org: failed to get record list from http://freedns.afraid.org/api/?action=getdyndns&v=2&sha=<credentials>: ERROR: Could not authenticate.
```
I think your forgot a new line.

rhansen:
> I think your forgot a new line.

That's intentional. The reason for the failure to get the record list was "ERROR: Could not authenticate". Separate lines implies two errors for two different actions, but this is a single error for a single action.

rhansen:
I'm going to merge this tomorrow unless there are any objections.

PR #186, opened 2020-06-21
Add is_ipv4() and ipv4_match() functions

dkerr64:
Current tests for IPv4 address are flawed... for example they match values >255 in an address.  To make the code more readable I have created two functions to;

1. test that a string is a valid IPv4 address and nothing more and
2. extract the first valid IPv4 address from a longer string or paragraph.

This matches the equivalent IPv6 functions (though the `ipv6_match()` function should be re-written as a single regex - outside scope of this PR).

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -949,7 +949,7 @@ sub update_nics {
                         if !$daemon || opt('verbose');
                     next;
                 }
-                if ($ip !~ /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/) {
+                if (!is_ipv4($ip)) {
                     if (!ipv6_match($ip)) {
                         warning("malformed IP address (%s)", $ip);
                         next;
@@ -1901,7 +1901,7 @@ sub check_value {
 
     } elsif ($type eq T_IP) {
         if (!ipv6_match($value)) {
-            return undef if $value !~ /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/;
+            return undef if !is_ipv4($value);
         }
     }
     return $value;
@@ -2289,8 +2289,7 @@ sub get_ip {
     }
     if (defined($ip)) {
         # no need to parse $reply
-    } elsif ($reply =~ /^.*?\b(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})\b.*/is) {
-        $ip = $1;
+    } elsif ($ip = ipv4_match($reply)) {
         $ip = un_zero_pad($ip);
         $ip = filter_local($ip) if opt('fw-banlocal', $h);
     } elsif ($ip = ipv6_match($reply)) {
@@ -2305,6 +2304,23 @@ sub get_ip {
 
     debug("get_ip: using %s, %s reports %s", $use, $arg, define($ip, "<undefined>"));
     return $ip;
+}
+
+######################################################################
+## is_ipv4() validates if string is valid IPv4 address and only
+## a valid IPv4 address (no preceding or trailing spaces/characters)
+######################################################################
+sub is_ipv4 {
+    my ($value) = @_;
+    return (length($value // '') != 0) && ($value eq (ipv4_match($value) // ''));
+}
+
+######################################################################
+## ipv4_match() extracts the first valid IPv4 address from given string
+######################################################################
+sub ipv4_match {
+    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
+    return $1;
 }
 
 ######################################################################


ddclient
```diff
@@ -2307,6 +2306,24 @@ sub get_ip {
...
+    return $value =~ /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\z/ai;
+}
+
+######################################################################
+## ipv4_match() extracts the first valid IPv4 address from given string
+######################################################################
+sub ipv4_match {
+    my ($value) = @_;
+    my $ip = undef;
+    $ip = $1 if $value =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
```
SuperSandro2000, line 2323: This should reuse ``is_ipv4``.

dkerr64: It can't.  At least I don't think it can.  The functions do two different things.  `is_ipv4` returns true/false and tests that a string exactly equals a valid IPv4 address... no text preceding or after the IPv4 in the string being tested.   `ipv4_match` on the other hand extracts the first valid IPv4 address it finds within a string (which could be multi-line, like output from `ifconfig`).  I'm not sure I like the name of the function but I chose it for consistency with existing `ipv6_match` function.

`ipv4_match` could be used instead of `is_ipv4` depending on what you want to do, but I don't think you can implement one function by calling the other.

dkerr64: You could do it the other way around... make `is_ipv4` use `ipv4_match`...
```
sub is_ipv4 {
    my ($value) = @_;
    return $value eq ipv4_match($value);
}
```
The downside is it adds a string compare, but the upside is that the regex is only in the one place.  I like this, What do you think @SuperSandro2000 

David

dkerr64: But would have to test for non-empty string as well.

SuperSandro2000: I think it would be better if we have the regex only once. I am not to deep into perl style but some global const (``IPV4_REGEX``) could work in other languages.

I did a sloppy read and thought you would duplicated code. Sorry about that.

dkerr64: Optimized I think the most efficient is...
```
sub is_ipv4 {
    my ($value) = @_;
    return (length($value) != 0) && ($value eq ipv4_match($value));
}
```
So test for empty string before going into the regex match.

dkerr64: > I did a sloppy read and thought you would duplicated code. Sorry about that.

No need to apologize... you prompted me to think about it and come up with a better way.  Maybe not more efficient, but definitely more readable.  Now I want to do the same to the IPv6 versions (as a separate PR after this one is merged)

ddclient
```diff
@@ -2307,6 +2306,25 @@ sub get_ip {
...
+    return $ip;
```
rhansen, line 2325: Similarly this will log a warning if `$value` is `undef`. How about this:

```suggestion
    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
    return $1;
```

ddclient
```diff
@@ -2307,6 +2306,25 @@ sub get_ip {
     return $ip;
 }
 
+######################################################################
+## is_ipv4() validates if string is valid IPv4 address and only
+## a valid IPv4 address (no preceding or trailing spaces/characters)
+######################################################################
+sub is_ipv4 {
+    my ($value) = @_;
+    return (length($value) != 0) && ($value eq ipv4_match($value));
```
rhansen, line 2315: If `$value` is `undef`, or if there's no IPv4 address in `$value`, then this will log a warning. How about this instead:

```suggestion
    return (($value // '') ne '') && ($value eq (ipv4_match($value) // ''));
```

dkerr64: I read that testing for length equal zero is more efficient than a string compare to test for an empty string.  Also, should not be necessary to test for undefined after the && as it will never get there.  So I suggest the following.
```
return (length($value // '') != 0) && ($value eq ipv4_match($value));
```

And, btw, thanks for pointing this out...  I have turned on warnings in my perl test script so that I catch things like this.

ddclient
```diff
@@ -2307,6 +2306,25 @@ sub get_ip {
...
+######################################################################
+sub is_ipv4 {
+    my ($value) = @_;
+    return (length($value // '') != 0) && ($value eq ipv4_match($value));
+}
+
+######################################################################
+## ipv4_match() extracts the first valid IPv4 address from given string
+## IPv4 address may be in form nn.nn.nn.nn or nn.nn.nn.nn/nn but the
+## function only returns the IP address without the mask.
```
rhansen, line 2321: Why do we need this? It makes the regex difficult to understand.

ddclient
```diff
@@ -2307,6 +2306,25 @@ sub get_ip {
     return $ip;
 }
 
+######################################################################
+## is_ipv4() validates if string is valid IPv4 address and only
+## a valid IPv4 address (no preceding or trailing spaces/characters)
+######################################################################
+sub is_ipv4 {
+    my ($value) = @_;
+    return (length($value // '') != 0) && ($value eq ipv4_match($value));
```
rhansen, line 2315: `ipv4_match` returns `undef` if there is no match. If that return value is directly compared with `$value` we'll get a warning.

```suggestion
    return (length($value // '') != 0) && ($value eq (ipv4_match($value) // ''));
```

ddclient
rhansen: > I read that testing for length equal zero is more efficient than a string compare to test for an empty string.

That was true in older versions of Perl, but I don't know if it's still true—I couldn't find any recent discussion about it. Even if `eq ''` is slower than a length check, using a length check feels like a premature micro-optimization.

In general I prefer to err on the side of human readability (and thus maintainability) rather than performance, unless performance is critical. In this case I'm OK with it either way; a length check is idiomatic enough that it won't surprise anyone.

ddclient
dkerr64: Yes.  For example th e output from `ip -4 -o addr show dev ens160`
```
 2: ens160    inet 192.168.17.2/24 brd 192.168.17.255 scope global noprefixroute ens160\       valid_lft forever preferred_lft forever
```
And the output from `ifconfig ens160 | grep "inet "`
```
inet 192.168.17.2  netmask 255.255.255.0  broadcast 192.168.17.255
```
And I want to be able to parse out the IPv4 address from either.

rhansen: What was wrong with the original regular expression? A digit followed by `/` matches `\b`, so there should be no need to explicitly look for a CIDR mask:

```shell
$ perl -Mstrict -w -e '
  my $x = " 2: ens160    inet 192.168.17.2/24 brd 192.168.17.255 scope global noprefixroute ens160";
  if ($x =~ /\b((?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai) {
    printf("%s\n", $1);
  } else {
    printf("no match\n");
  }
'
192.168.17.2
```

rhansen: Here's a more compact RE:  `/\b((?:(25[0-5]|(?:2[0-4]|1\d|[1-9])?\d)\.){3}(?2))\b/a`

dkerr64: So we have to decide...

1. For CIDR /xx do we care what xx is?  My regex checked that it was in the range of 0 to 32.
2. Do we allow for leading zeros within the IP... should we accept 192.168.100.001 for example or only 192.168.100.1  again my regex accepted the leading zeros.

ddclient currently has code to remove the leading zeros, but only after validating the IPv4.  And we should validate a IPv4 provided by the user where leading zeros may be provided even though not strictly legal.  So I think we need to allow for (2) but as the CIDR is irrelevant we can dump that part of the regex.

rhansen: > So we have to decide...
> 
>   1. For CIDR /xx do we care what xx is?  My regex checked that it was in the range of 0 to 32.

Nope, we don't care. It has no relevance to ddclient.

> 
>   2. Do we allow for leading zeros within the IP... should we accept 192.168.100.001 for example or only 192.168.100.1  again my regex accepted the leading zeros.

It's been ages since I've seen anyone zero-pad, and on some systems a leading 0 means octal. I guess we might as well continue to accept leading zeros given that we have historically accepted them. We can remove support for leading zeros in a follow-up commit if desired.

For the new `usev4` family of options I would prefer to reject leading zeros, if doing so wouldn't make the code overly complicated.

dkerr64: I have simplified the regex.  Yes makes sense to reject leading zeros in the `usev4` version.

rhansen:
Thank you very much for putting this together!

PR #187, opened 2020-06-23
Use the `//` operator to simplify code

rhansen:


ddclient
```diff
@@ -1708,10 +1708,7 @@ sub opt {
     my $v = shift;
     my $h = shift;
     return $config{$h}{$v} if defined($h) && defined($config{$h}{$v});
-    return $opt{$v}        if defined $opt{$v};
-    return $globals{$v}    if defined $globals{$v};
-    return default($v)     if defined default($v);
-    return undef;
+    return $opt{$v} // $globals{$v} // default($v);
```
SuperSandro2000, line 1711: Does this still return undef?

rhansen: Yes. Note that:

```perl
    return default($v)     if defined default($v);
    return undef;
```

is equivalent to:

```perl
    return default($v);
```

ddclient
```diff
@@ -1022,7 +1022,7 @@ sub write_cache {
     ## construct the cache file.
     my $cache = "";
     foreach my $h (sort keys %cache) {
-        my $opt = join(',', map { "$_=" . define($cache{$h}{$_},'') } sort keys %{$cache{$h}});
+        my $opt = join(',', map { "$_=" . ($cache{$h}{$_} // '') } sort keys %{$cache{$h}});
```
SuperSandro2000, line 1025:
```suggestion
        my $opt = join(',', map { "$_=" . $cache{$h}{$_} // '' } sort keys %{$cache{$h}});
```

rhansen: No, because `.` has higher precedence than `//`. Removing the parentheses is equivalent to:

```suggestion
        my $opt = join(',', map { ("$_=" . $cache{$h}{$_}) // '' } sort keys %{$cache{$h}});
```

PR #191, opened 2020-06-26
Turn `fw-banlocal` into a no-op and mark it as deprecated

rhansen:
`fw-banlocal` is problematic:
  * There's not much point to it. Regardless of whether it is enabled, the end result is a DNS record that is not being updated to a useful value. It does cause a warning to be logged, but because it is not enabled by default it doesn't help the poor user who is trying to figure out why they can't reach their machine. By the time they realize that enabling this option would have saved them hours of troubleshooting, they no longer need to enable it because they already know what the problem is.
  * It's a misnomer: `fw-banlocal` doesn't just filter out local IP addresses from `use=fw`, it also filters them out of all other address sources except `use=ip`.
  * It doesn't filter out local IPv6 addresses.
  * The resulting warning ("unable to determine IP address") is misleading.

We might want to add a warning whenever a non-global address is discovered (along with an option to silence the warning), but that should be done in a future commit if at all.

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -336,7 +336,6 @@ my %variables = (
         'web-skip'            => setv(T_STRING,1, 0, 1, '',                   undef),
         'fw'                  => setv(T_ANY,   0, 0, 1, '',                   undef),
         'fw-skip'             => setv(T_STRING,1, 0, 1, '',                   undef),
-        'fw-banlocal'         => setv(T_BOOL,  0, 0, 1, 0,                    undef),
         'fw-login'            => setv(T_LOGIN, 1, 0, 1, '',                   undef),
         'fw-password'         => setv(T_PASSWD,1, 0, 1, '',                   undef),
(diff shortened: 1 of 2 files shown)


rhansen:
cc @dkerr64

dkerr64:
Yes definitely remove it.

ddclient
```diff
@@ -2299,10 +2284,8 @@ sub get_ip {
         # no need to parse $reply
```
dkerr64, line 2284: I think we should change this such that `ip=<addr>` is processed through `ipv4_match` (and `ipv6_match`) in order to validate that the hard coded IP address provided by the user is in fact valid... so where we read in the ip address above place it onto `$reply` and then go through the following code to set `$ip`.

rhansen: Agreed, though that's out of scope for this pull request. I'll send a follow-up PR.

PR #192, opened 2020-06-26
Cleanup ipv6_match, ipv4_match, add ipv6_match_gua

dkerr64:
Replace `ipv6_match` function with one based on regex (from _Regular Expressions Cookbook by J.Goyvaerets and S.Levithan_ published by O'Reilly) and add `ipv6_match_gua` function.  These and an updated `ipv4_match` will accept IP addresses with embedded leading zeros (e.g. 192.168.001.001) but return with the zeros removed (e.g. 192.168.1.1) thus making the `un_zero_pad` function redundant so it is removed.

`is_ipv4` and `is_ipv6` functions remain however these should be redundant as the match functions can be used for the same end result.  I have removed all calls to these but not the functions themselves pending further review.


dkerr64:
Note that this touches same parts of the code as #191 so likely will have to do manual merge if one is merged before the other.

dkerr64:
> Note that this touches same parts of the code as #191 so likely will have to do manual merge if one is merged before the other.

conflict resolved.

ddclient
```diff
@@ -2298,57 +2281,72 @@ sub get_ip {
...
-## ipv4_match() extracts the first valid IPv4 address from given string
+## ipv4_match() extracts the first valid IPv4 address from given string.
+## Accepts leading zeros in the address but removes them in returned value
 ######################################################################
 sub ipv4_match {
-    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
-    return $1;
+    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3} ## first three bytes
+                         (?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/xai;  ## last one
+    return ($1 // '') =~ s/\b0+\B//gr;                                     ## remove leading zeros
```
rhansen, line 2300: The `/r` flag is only available in Perl 5.14+ so it can't be used. (We support Perl 5.10.1 and later.)

ddclient
```diff
@@ -2298,57 +2281,72 @@ sub get_ip {
...
-            # or one double colon
-            my $count_double_colon = () = $parsed =~ /::/g;
-            if ($count_double_colon != 1) {
-                next
-            }
-        }
-        return $parsed;
-    }
-    return;
+    (shift // '') =~ /(?<![:.\w])                                     ## Negative Look behind for word boundry
```
rhansen, line 2319: Why not `\b`?

ddclient
```diff
@@ -2298,57 +2281,72 @@ sub get_ip {
...
+                      ( (?:[0-9A-F]{1,4}:){7}[0-9A-F]{1,4}            ## Standard format
+                       |(?=(?:[0-9A-F]{0,4}:){0,7}[0-9A-F]{0,4}       ## OR compressed with at most 7 colons
+                        (?![:.\w]))                                   ##   and negative lookahead for boundry
+                        (?:(?:[0-9A-F]{1,4}:){1,7}|:)(?:(?::[0-9A-F]{1,4}){1,7}|:)  ## and only 1 double-colon
+                       |(?:[0-9A-F]{1,4}:){7}:                        ## OR compressed with 8 colons (double at end)
+                       |:(?::[0-9A-F]{1,4}){7}                        ## OR compressed with 8 colons (double at front)
+                      )                                               ## End of first capture group
+                      (?![:.\w])/xai;                                 ## Negative lookahead for boundry
```
rhansen, line 2327: This is complicated enough that I don't feel comfortable merging this without unit tests.

ddclient
```diff
@@ -4256,7 +4254,7 @@ sub nic_nsupdate_update {
         my $zone       = $config{$h}{'zone'};
         my $ip         = $config{$h}{'wantip'};
         my $recordtype = '';
-        if (is_ipv6($ip)) {
+        if (ipv6_match($ip)) {
```
rhansen, line 4257: Why make this change? `is_ipv6` looks more correct to me.

dkerr64:
@rhansen Thank you for the review. I will go over each comment and respond.

ddclient
dkerr64: Well that's a bummer.  And I was so proud of myself as its rather elegant.  Oh well, have to create a local variable.

ddclient
```diff
@@ -2283,9 +2266,9 @@ sub get_ip {
...
         # no need to parse $reply
     } elsif ($ip = ipv4_match($reply)) {
-        $ip = un_zero_pad($ip);
+        # got an IPv4 address
     } elsif ($ip = ipv6_match($reply)) {
-        $ip = un_zero_pad($ip);
+        # got a IPv6 address
     } else {
         warning("found neither ipv4 nor ipv6 address");
     }
```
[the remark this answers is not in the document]
dkerr64: Yes, nice.  I just wanted to resolve the merge conflict and was rather unhappy with how it looked.  I expected there was a better way to do it.  Thanks for the much better way.

ddclient
dkerr64: Thats a good question.  I went with what the book said and didn't test alternatives (interestingly the book uses `\b` for IPv4 but not for IPv6).  I'll take another look.

ddclient
dkerr64: So thats the debate I'd like to have.  Is it worthwhile having `is_ipv4` and `is_ipv6` as separate functions when using the `ipvN_match` functions can accomplish much the same thing?  It is certainly easier reading to come across a test for `is_something` but then should we have these functions just for readability?

The issue that prompts me to ask for the debate is that I updated the `ipvN_match` functions to strip out embedded leading zeros.  Which means that if you test `is_ipv4("192.168.001.001")` then it is going to return false.  As long as we are comfortable with this then I would keep the functions and use them -- especially as 99% of the time what we're testing came from `get_ip` -- inside of which we should remove any embedded leading zeros.

ddclient
dkerr64: Okay, so what's actually going on here is that the regex is not looking for a word break, rather it is looking for (accepting) any characters that are valid within a IPv6 address... in this case `[:.\w]` (colon, period and a word character) now this is probably not exactly right for what we're doing -- I only test for standard and compressed formats, I do not accept mixed (IPv4 as the last 32 bits of an IPv6 address), so accepting a period is wrong (I suspect the book used this so that their regex examples were similar when accepting mixed format.  And `\w` also matches things like underscore which would be wrong also (the regex does not extract `::1` from `::1_` because the underscore is a word character).

So, for our purposes where we are looking only for a standard or compressed format IPv6 I think the lookahead / lookbehind should be for `[:0-9A-F]` case insensitive.  This is working for my list of test IPv6's.

Thanks for making me look at this.

ddclient
dkerr64: How would you like to do that?  There are various places where I have found lists of valid and invalid IPv6 addresses which can be used for test purposes.  The lists are not very good however -- they do test for valid formatting (number of colons, colons in wrong places, etc) but don't include examples of GUA, ULA and LLA addresses which I think a good test bucket should include, certainly for our purposes.

ddclient
dkerr64: I've done further testing this morning and concluded that the book's lookahead/behind is actually correct.  If I omit the period then it matches `1111::123` from the string `1111::123.123.123.123` when it should reject the whole string (it is a mixed IPv6 rather than a standard or compressed IPv6).  Same for `[0-9A-F]` and `\w`.  It needs to be `\w` else the it matches `1111::123` from `1111::123_123_123_123` which again is wrong.  So I'm back to using what the book recommends.

ddclient
```diff
@@ -942,14 +942,14 @@ sub update_nics {
             if (exists $iplist{$use}{$arg_ip}{$arg_fw}{$arg_if}{$arg_web}{$arg_cmd}) {
                 $ip = $iplist{$use}{$arg_ip}{$arg_fw}{$arg_if}{$arg_web}{$arg_cmd};
             } else {
-                $ip = get_ip($use, $h) // '';
-                if (!$ip) {
+                my $reply = get_ip($use, $h) // '';
+                if (!$reply) {
```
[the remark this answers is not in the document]
dkerr64: Reverted

ddclient
```diff
@@ -942,14 +942,14 @@ sub update_nics {
...
                         if !$daemon || opt('verbose');
                     next;
                 }
-                if (!is_ipv4($ip)) {
-                    if (!ipv6_match($ip)) {
+                if (!($ip = ipv4_match($reply))) {
+                    if (!($ip = ipv6_match($reply))) {
                         warning("malformed IP address (%s)", $ip);
                         next;
                     }
```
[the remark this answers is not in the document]
dkerr64: Done

ddclient
```diff
@@ -1901,7 +1901,9 @@ sub check_value {
 
     } elsif ($type eq T_IP) {
         if (!ipv6_match($value)) {
-            return undef if !is_ipv4($value);
+            return undef if !($value = ipv4_match($value));
+        } else {
+            $value = ipv6_match($value);
         }
```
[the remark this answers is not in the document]
dkerr64: Done

ddclient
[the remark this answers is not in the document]
dkerr64: Done.  This does require that `$ip` is set to `undef` when first defined (which it is) and also I set the value of a manually entered ip address into `$reply` rather than into `$ip` so that we will validate (and remove any embedded leading zeros) any address provided.

ddclient
```diff
@@ -2298,57 +2281,72 @@ sub get_ip {
...
 ######################################################################
-## ipv4_match() extracts the first valid IPv4 address from given string
+## ipv4_match() extracts the first valid IPv4 address from given string.
+## Accepts leading zeros in the address but removes them in returned value
 ######################################################################
 sub ipv4_match {
-    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
-    return $1;
+    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3} ## first three bytes
+                         (?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/xai;  ## last one
```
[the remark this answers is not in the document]
dkerr64: I made the change.  Thanks for the suggestion.

ddclient
dkerr64: I decided to leave `is_ipv4` and `is_ipv6` in and use them (so I changed this back to how it was).

ddclient
rhansen: > How would you like to do that?

You'll have to wait for PR #161 to be merged.  Until then, split this into two PRs: one that doesn't require tests and one that does.

ddclient
```diff
@@ -2183,7 +2156,7 @@ sub get_ip {
     $arg = '' unless $arg;
 
     if ($use eq 'ip') {
-        $ip  = opt('ip', $h);
+        $reply  = opt('ip', $h);
```
rhansen, line 2159: Values that aren't exactly an IP address should be rejected:

```suggestion
        $ip = opt('ip', $h);
        $ip = undef if !is_ipv4($ip) && !is_ipv6($ip);
```

ddclient
```diff
@@ -2298,57 +2264,78 @@ sub get_ip {
...
 ######################################################################
-sub ipv4_match {
-    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
-    return $1;
+sub ipv4_extract{
+    (shift // '') =~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3} ## first three bytes
+                         (?&octet))\b/xa;                               ## last one (without period)
+    my $ip = ($1 // '');
+    $ip =~ s/\b0+\B//g;                                                ## remove leading zeros
+    return $ip;
```
rhansen, line 2285: This function should return either an IP address or `undef`, never an empty string.

```suggestion
    if ((shift // '') !~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?&octet))\b/a) {
        return undef;
    }
    (my $ip = $1) =~ s/\b0+\B//g;  # remove leading zeros
    return $ip;
```

ddclient
dkerr64: Updated and added a warning message.

ddclient
dkerr64: Thanks for the suggestion, change made.

ddclient
```diff
@@ -2184,6 +2157,10 @@ sub get_ip {
 
     if ($use eq 'ip') {
         $ip  = opt('ip', $h);
+        if (!is_ipv4($ip) && !is_ipv6($ip)) {
+            warning("'%s' is not a valid IPv4 or IPv6 address, ignoring value.",$ip);
```
rhansen, line 2161: "Ignoring value" might be misleading depending on what the caller does with `undef`.

```suggestion
            warning("'%s' is not a valid IPv4 or IPv6 address", $ip);
```

ddclient
```diff
@@ -2298,57 +2268,81 @@ sub get_ip {
...
-## ipv4_match() extracts the first valid IPv4 address from given string
+## extract_ipv4() extracts the first valid IPv4 address from given string.
+## Accepts leading zeros in the address but removes them in returned value
 ######################################################################
-sub ipv4_match {
-    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
-    return $1;
+sub extract_ipv4{
+    if ((shift // '') !~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}
+                         (?&octet))\b/xa) {
```
rhansen, line 2286: The regex is short enough that you can remove the line break and it'll still fit in <100 columns:

```suggestion
    if ((shift // '') !~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?&octet))\b/a) {
```

If you want to split it across multiple lines so that you can embed comments (I don't think they're necessary for this regex), it would probably be more readable to store the regex in a variable:

```suggestion
    my $re = qr/\b(
            (?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}  # first three octets
            (?&octet)                                       # last octet
        )\b/xa;
    if ((shift // '') !~ $re) {
```

ddclient
```diff
@@ -2298,57 +2268,81 @@ sub get_ip {
...
-    my @values = split('\n', $content);
-    foreach my $val (@values) {
-        next unless $val =~ /((:{0,2}[A-F0-9]{1,4}){0,7}:{1,2}[A-F0-9]{1,4})/ai; # invalid char
-        my $parsed = $1;
+## extract_ipv6() extracts the first valid IPv6 address from given string
+## IPv6 must be in standard or compressed format.
+## Mixed IPv6/IPv4 not supported.
+## Accepts leading zeros in the address but removes them in returned value
+######################################################################
+sub extract_ipv6{
```
rhansen, line 2308: Move this to a second pull request.

ddclient
```diff
@@ -2298,57 +2268,81 @@ sub get_ip {
...
-            }
-        }
-        return $parsed;
+######################################################################
+## extract_ipv6_gua() extracts the first valid IPv6 Global Unicast Address (GUA)
+## from given string. IPv6 must be in standard or compressed format.
+## Mixed IPv6/IPv4 not supported.
+## Accepts leading zeros in the address but removes them in returned value
+######################################################################
+sub extract_ipv6_gua {
```
rhansen, line 2330: Move this to a third pull request.

dkerr64:
Splitting this into three.  Withdrawing this PR.

PR #193, opened 2020-06-27
Delete unused `config` entry

rhansen:


SuperSandro2000:
I really want unit tests before merging such stuff to be sure that nothing broke.

rhansen:
I'm not sure that's feasible right now. The config logic is complicated, scattered, and inconsistent; writing tests will first require a major refactoring.

PR #194, opened 2020-06-27
Rename `ipv4_match`, `ipv6_match` to `ipv4_extract`, `ipv6_extract`

rhansen:
"Match" only implies a boolean return value. While these functions can be used in boolean context, "extract" more closely matches their intended purpose.

cc dkerr64

### DIFF
--- a/ddclient
+++ b/ddclient
@@ -949,7 +949,7 @@ sub update_nics {
                     next;
                 }
                 if (!is_ipv4($ip)) {
-                    if (!ipv6_match($ip)) {
+                    if (!extract_ipv6($ip)) {
                         warning("malformed IP address (%s)", $ip);
                         next;
                     }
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Not sure if we should name it ``extract_ipv4`` and ``extract_ipv6``.

dkerr64:
I agree we should rename.  I think it should be done as part of #192 rather than its own PR.  If you prefer a separate PR then please merge so I can then update #192 to match.  I don't care if it is `ipvN_extract` or `extract_ipvN` but we will have to also have a `ipv6_gua_extract` or `extract_ipv6_gua` as well (depending on what you decide) so keep that in mind.

Thanks.

SuperSandro2000:
@dkerr64 do you think rebasing will easy for you or requires more work? I don't want to make your work harder.

dkerr64:
> @dkerr64 do you think rebasing will easy for you or requires more work? I don't want to make your work harder.

It should be easy enough.

SuperSandro2000:
> Not sure if we should name it extract_ipv4 and extract_ipv6.

@rhansen

rhansen:
Renamed to `extract_ipvx`

PR #195, opened 2020-06-29
Move per-service variables to the service definitions

rhansen:
Most of the entries in `%variables` are only used once. Move them down to their respective service definitions.

Also:
  * Use idiomatic hash merging instead of `merge()`.
  * Sort the variables by name.


SuperSandro2000:
My quick dirty smoke test already found this bug:
```
Can't use an undefined value as a HASH reference at ./ddclient line 678.
```

rhansen:
Done. I'm scratching my head over this one; I'm not sure why I didn't see it before and I'm not sure what was wrong. Maybe I hit a syntax corner case? I just added extra parentheses and it fixed it.

SuperSandro2000:
I don't know but for that reason we are adding tests.

PR #196, opened 2020-06-29
Cleanup extract ipv4 and is ipv6 and usage

dkerr64:
Splitting PR #192 into multiple parts.  This is the simplification of extract_ipv4, is_ipv6 and usage of them.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -2212,6 +2212,10 @@ sub get_ip {
 
     if ($use eq 'ip') {
         $ip  = opt('ip', $h);
+        if (!is_ipv4($ip) && !is_ipv6($ip)) {
+            warning("'%s' is not a valid IPv4 or IPv6 address", $ip);
+            $ip = undef;
+        }
         $arg = 'ip';
 
     } elsif ($use eq 'if') {
--- a/ddclient.in
+++ b/ddclient.in
@@ -984,12 +984,6 @@ sub update_nics {
                         if !$daemon || opt('verbose');
                     next;
                 }
-                if (!is_ipv4($ip)) {
-                    if (!extract_ipv6($ip)) {
-                        warning("malformed IP address (%s)", $ip);
-                        next;
-                    }
-                }
                 $iplist{$use}{$arg_ip}{$arg_fw}{$arg_if}{$arg_web}{$arg_cmd} = $ip;
             }
             $config{$h}{'wantip'} = $ip;
--- a/ddclient.in
+++ b/ddclient.in
@@ -2342,7 +2342,7 @@ sub is_ipv4 {
 ## extract_ipv4() extracts the first valid IPv4 address from the given string
 ######################################################################
 sub extract_ipv4 {
-    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
+    (shift // '') =~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?&octet))\b/a;
     return $1;
 }
 
--- a/ddclient.in
+++ b/ddclient.in
@@ -2181,26 +2181,7 @@ sub geturl {
     $reply =~ s/\r//g if defined $reply;
     return $reply;
 }
-######################################################################
-## un_zero_pad
-######################################################################
-sub un_zero_pad {
-    my $in_str  = shift(@_);
-    my @out_str = ();
 
-    if ($in_str eq '0.0.0.0') {
-        return $in_str;
-    }
-
-    foreach my $block (split /\./, $in_str) {
-        $block =~ s/^0+//;
-        if ($block eq '') {
-            $block = '0';
-        }
-        push @out_str, $block;
-    }
-    return join('.', @out_str);
-}
 ######################################################################
 ## get_ip
 ######################################################################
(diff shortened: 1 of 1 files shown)


ddclient
```diff
@@ -2345,6 +2321,7 @@ sub extract_ipv6 {
                 next
             }
         }
+        $parsed =~ s/\b0+\B//g; ## remove embedded leading zeros
```
rhansen, line 2324: This worries me a bit because I believe leading zeros are fairly common in real-world IPv6 addresses. I don't want `is_ipv6` to reject a legitimate IPv6 address unless the format is highly unusual. Maybe leave this out for now?

dkerr64: Well, I think that should be resolved in `is_ipv6` and not in `extract_ipv6`.  The purpose of this function is to obtain the IPv6 address from a larger body of text (which could in fact be nothing more than an IP address) for use elsewhere in the program... e.g. to send up to a dynamic DNS hosting site.  I think that `extract_ipv6` should return the _best_ format for that purpose and I think embedded leading zeros should be removed.

Now I do agree that this is problematic for `is_ipv6` which now fails on a perfectly fine IPv6 address.  This really only matters for an address manually set by the user in `ip=<address>` or `ipv6=<address>`, if you remember you had me switch from using `extract_ipv6` over there to using `is_ipv6`.   So I would rather fix this by either running manually entries through extract first, or  changing `is_ipv6` to accept addresses with embedded leading zeros.  I think I prefer updating `is_ipv6` to accept them -- let me propose a change to that and see what you think.

Whatever we do we should do the same for IPv4.

rhansen: > I think I prefer updating is_ipv6 to accept them

Yup, I agree.

> Whatever we do we should do the same for IPv4.

I disagree. I'm OK with rejecting leading zeros for IPv4 because they're extremely rare nowadays and can be ambiguous (some systems interpret leading zeros in an IPv4 address as signifying octal). For IPv6, leading zeros does not cause ambiguity; it's just not as compact.

dkerr64:
Now I am stuck until #206 is addressed.

ddclient
dkerr64: Take a look at the change I just committed to accept embedded leading zeros within the IPv6 address.

dkerr64: > Take a look at the change I just committed to accept embedded leading zeros within the IPv6 address.  This is another place that the`/r` regex modifier would have been useful:-)

dkerr64: > I disagree. I'm OK with rejecting leading zeros for IPv4 because they're extremely rare nowadays and can be ambiguous (some systems interpret leading zeros in an IPv4 address as signifying octal). For IPv6, leading zeros does not cause ambiguity; it's just not as compact.

Fine with me.

ddclient
```diff
@@ -2297,29 +2267,36 @@ sub get_ip {
...
-## is_ipv6() validates if string is valid IPv6 address
+## is_ipv6() validates if string is valid IPv6 address with no preceding
+## or trailing spaces/characters. Embedded leading zeros are accepted
+## but removed for comparison, so extract_ipv6 must remove them also.
 ######################################################################
 sub is_ipv6 {
     my ($value) = @_;
-    # This little gem from http://home.deds.nl/~aeron/regex/
-    return $value =~ /^(((?=.*(::))(?!.*\3.+\3))\3?|([\dA-F]{1,4}(\3|:\b|$)|\2))(?4){5}((?4){2}|(((2[0-4]|1\d|[1-9])?\d|25[0-5])\.?\b){4})\z/ai;
+    return (length($value // '') != 0) && ((extract_ipv6($value) // '') eq (($value =~ s/\b0+\B//g)?$value:$value));
```
rhansen, line 2299: Interesting approach, and a clever way around the lack of `/r`. A concern I have is whether its understandable by a newcomer examining the code for the first time.

The following are other ways around `/r`, but I don't think they're any more readable:

```perl
do { $value =~ s/\b0+\B//g; $value }
scalar($value =~ s/\b0+\B//g, $value)
```

The following requires more lines of code, but maybe the readability trade-off is worthwhile?

```perl
sub _extract_ipv6 {
    # copy of current extract_ipv6
}

sub extract_ipv6 {
    my $ip = _extract_ipv6(shift) or return undef;
    $ip =~ s/\b0+\B//g;
    return $ip;
}

sub is_ipv6 {
  my ($value) = @_;
  return ($value // '') ne '' && $value eq (_extract_ipv6($value) // '');
}
```

The regular expression version could look like this:

```perl
my $ipv6_regex = qr/.../;

sub extract_ipv6 {
    (shift // '') =~ /\b($ipv6_regex)\b/ or return undef;
    (my $ip = $1) =~ s/\b0+\B//g;
    return $ip;
}

sub is_ipv6 {
    return (shift // '') =~ /^$ipv6_regex$/;
}
```

dkerr64: I actually like your last one.... move the regex into a global variable and then use it in both functions.

ddclient
```diff
@@ -2297,29 +2267,35 @@ sub get_ip {
...
-## extract_ipv4() extracts the first valid IPv4 address from the given string
+## extract_ipv4() extracts the first valid IPv4 address from given string.
+## Accepts leading zeros in the address but removes them in returned value
 ######################################################################
 sub extract_ipv4 {
-    (shift // '') =~ /\b((?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\b/ai;
-    return $1;
+    if ((shift // '') !~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?&octet))\b/a) {
+        return undef;
+    }
```
rhansen, line 2286: A bit more compact, if you prefer:

```suggestion
    (shift // '') =~ /\b((?:(?<octet>25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?&octet))\b/a)
        or return undef;
```

ddclient
rhansen: How about we go with what you've got for now, and in PR #198 you can switch to the last one?

SuperSandro2000:
I would prefer it if we could rebase PR instead of merging master.

ddclient
dkerr64: > How about we go with what you've got for now, and in PR #198 you can switch to the last one?

Yes, because that last one is going to take some testing.

dkerr64:
> I would prefer it if we could rebase PR instead of merging master.

I'm not familiar with rebase so went with what I know.  I'll try and learn how to properly rebase.

ddclient
dkerr64: I have added that in too.... you had an extra close bracket at end of line (syntax error) that needed to be removed.

rhansen:
I rebased onto latest `master` and split the individual changes into separate commits.

rhansen:
Oops, looks like you did too. :)

dkerr64:
> I rebased onto latest `master` and split the individual changes into separate commits.

Thank you... you and I did a rebase at the same time I think.  But end result is the same !

dkerr64:
If you could merge this into master please... I'm going to focus on unit tests for the other two PRs to get those ready for merging.

PR #208, opened 2020-06-30
Fix shebang

SuperSandro2000:
This is probably not the nicest method but it fixes the shebang so that you can run the script normally or with perl.
I also ignored the changed ``tap-driver.sh`` cause it is annoying when changing branches a lot.

I got inspiration from lcov https://github.com/linux-test-project/lcov/commit/74bae96e8ef724eb9dbdf126adad17505375e149#diff-2066c4d7515c02785d28a6ab3a57ace1R41 .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ release
 /autom4te.cache/
 /build-aux/install-sh
 /build-aux/missing
+/build-aux/tap-driver.sh
 /config.log
 /config.status
 /configure


rhansen:
I merged the shebang commit. Unfortunately the `tap-driver.sh` commit does nothing—Git doesn't support ignoring tracked files. The closest you can get is running:

```shell
git update-index --assume-unchanged -- build-aux/tap-driver.sh
```

We can, however, modify `autogen` to revert the changes to `tap-driver.sh` after `autoreconf` runs. I'll put together a PR.

PR #209, opened 2020-06-30
Add test for centos, use matrix for ubuntu

SuperSandro2000:
I think it is not running cause I changed it. I think if a member does such stuff the action should run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,3 +28,19 @@ jobs:
         run: make VERBOSE=1 check
       - name: distcheck
         run: make VERBOSE=1 distcheck
+
+  test-centos8:
+    runs-on: ubuntu-latest
+    container: centos:8
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: dnf --enablerepo=PowerTools install -y perl-version automake make perl-Test-Warnings
+      - name: autogen
+        run: ./autogen
+      - name: configure
+        run: ./configure
+      - name: check
+        run: make VERBOSE=1 check
+      - name: distcheck
+        run: make VERBOSE=1 distcheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,6 @@ jobs:
       - name: configure
         run: ./configure
       - name: check
-        run: make check
+        run: make VERBOSE=1 check
       - name: distcheck
-        run: make distcheck
+        run: make VERBOSE=1 distcheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,3 +60,19 @@ jobs:
         run: make VERBOSE=1 check
       - name: distcheck
         run: make VERBOSE=1 distcheck
+
+  test-fedora:
+    runs-on: ubuntu-latest
+    container: fedora
+    steps:
+      - uses: actions/checkout@v2
+      - name: install dependencies
+        run: dnf install -y perl-version automake make perl-Test-Warnings
+      - name: autogen
+        run: ./autogen
+      - name: configure
+        run: ./configure
+      - name: check
+        run: make VERBOSE=1 check
+      - name: distcheck
+        run: make VERBOSE=1 distcheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,23 +7,15 @@ on:
   pull_request:
     branches:
       - master
+
 jobs:
-  test-ubuntu-latest:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - name: install test dependencies
-        run: sudo apt-get install libtest-warnings-perl
-      - name: autogen
-        run: ./autogen
-      - name: configure
-        run: ./configure
-      - name: check
-        run: make check
-      - name: distcheck
-        run: make distcheck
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
@rhansen Do you think we should also test fedora or is centos enough for now?

.github/workflows/ci.yml
```diff
@@ -22,12 +26,16 @@ jobs:
         run: make check
       - name: distcheck
         run: make distcheck
-  test-ubuntu-oldest:
-    runs-on: ubuntu-16.04
+
+  test-centos:
+    runs-on: ubuntu-latest
+    container: centos
```
rhansen, line 32: [The documentation](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer) suggests that this should instead be:

```suggestion
    container:
      image: centos
```

Yet it seemed to run OK so `¯\_(ツ)_/¯`

.github/workflows/ci.yml
```diff
@@ -22,12 +26,16 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
+      - name: install ddclient dependencies
+        run: dnf install -y perl-version
       - name: install test dependencies
-        run: sudo apt-get install libtest-warnings-perl
+        run: dnf --enablerepo=PowerTools install -y autoconf automake make perl-Test-Warnings
```
rhansen, line 38: I'm not familiar with CentOS. Is this PowerTools repo required? I would like to test it on a CentOS install that is as vanilla as possible.

.github/workflows/ci.yml
```diff
@@ -22,12 +26,16 @@ jobs:
         run: make check
       - name: distcheck
         run: make distcheck
-  test-ubuntu-oldest:
-    runs-on: ubuntu-16.04
+
+  test-centos:
```
rhansen, line 30: Can we do a matrix of Docker images like this:

```yaml
  test-red-hat:
    strategy:
      matrix:
        image:
          - centos:latest
          - centos:6
          - fedora:latest
          - fedora:26
```

.github/workflows/ci.yml
```diff
@@ -7,9 +7,13 @@ on:
...
     branches:
       - master
+
 jobs:
-  test-ubuntu-latest:
-    runs-on: ubuntu-latest
+  test-ubuntu:
+    strategy:
+      matrix:
+        ubuntu: [latest, 16.04]
```
rhansen, line 15: Very cool!

rhansen, line 15: Nit:

```suggestion
        ubuntu:
          - latest
          - 16.04
```

.github/workflows/ci.yml
SuperSandro2000: Nope, cause they use different commands.

.github/workflows/ci.yml
SuperSandro2000: ``PowerTools - Available only for CentOS8, the PowerTools repository provides most of the developer tools. Disabled by default.`` https://wiki.centos.org/AdditionalResources/Repositories

For ``perl-Test-Warnings`` it is required and is fine to install.

SuperSandro2000:
I just noticed that this is Cenots 8 but Centos 6 has different commands to get the lib perl. I will add it with Fedora just to get it done for once.

.github/workflows/ci.yml
SuperSandro2000: I used https://centos.pkgs.org/8/centos-powertools-x86_64/perl-Test-Warnings-0.026-7.el8.noarch.rpm.html to get the required packages and instructions.

SuperSandro2000:
I would have wished to use yaml anchors for the repeating test commands but github actions is not supporting them at this point but they are planning it for the future.

.github/workflows/ci.yml
```diff
@@ -7,9 +7,15 @@ on:
...
       - master
+
 jobs:
-  test-ubuntu-latest:
-    runs-on: ubuntu-latest
+  test-ubuntu:
+    strategy:
+      matrix:
+        ubuntu:
+          - 16.04
```
rhansen, line 16: I believe YAML will treat this as a floating-point value rather than a string. But it looks like it works OK.

.github/workflows/ci.yml
```diff
@@ -22,12 +28,52 @@ jobs:
...
-  test-ubuntu-oldest:
-    runs-on: ubuntu-16.04
+
+  test-centos6:
+    runs-on: ubuntu-latest
+    container: centos:6
+    steps:
+      - uses: actions/checkout@v1
+      - name: enable special interest group
+        run: yum install centos-release-scl-rh
```
rhansen, line 38: Does this enable a 3rd party repository? I'd rather stick with the core repositories. It's OK if Test::Warnings is not installed, it'll just skip the warnings tests.

.github/workflows/ci.yml
SuperSandro2000: As long as it works we don't need to change it.

rhansen:
Apologies @SuperSandro2000, I thought you were out for the day and I force-pushed some changes that overwrote yours. Do you want to coordinate over IRC?

.github/workflows/ci.yml
SuperSandro2000: AS long as it is just for testing it is fine. Just makes our life easier.

PR #216, opened 2020-06-30
Explicit require version

SuperSandro2000:
because centos does not include it at all times. Right now we need to explicitly [install it for the tests to pass](https://github.com/ddclient/ddclient/blob/master/.github/workflows/ci.yml#L54). So I think it would be a good idea to also require it in automake just to be sure it is not forgotten.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -21,9 +21,12 @@ AX_PROG_PERL_VERSION([5.10.1], [],
   [AC_MSG_ERROR([Perl 5.10.1 or newer not found])])
 AC_SUBST([PERL])
 
-# Perl modules required to run ddclient. Core modules may be omitted;
-# they are assumed to always exist.
+# Perl modules required to run ddclient. Note: CentOS, RHEL, and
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -24,6 +24,7 @@ AC_SUBST([PERL])
 # Perl modules required to run ddclient. Core modules may be omitted;
 # they are assumed to always exist.
 m4_foreach_w([_m], [
+    version
```
rhansen, line 27:
```suggestion
    version=0.77
```

configure.ac
```diff
@@ -24,6 +24,7 @@ AC_SUBST([PERL])
 # Perl modules required to run ddclient. Core modules may be omitted;
 # they are assumed to always exist.
```
rhansen, line 25:
```suggestion
# Perl modules required to run ddclient. Note: CentOS, RHEL, and
# Fedora put some core modules in separate packages, and the perl
# package doesn't depend on all of them, so they can't be assumed to
# be available.
```

PR #218, opened 2020-07-01
Add redhat to ci

SuperSandro2000:


.github/workflows/ci.yml
```diff
@@ -76,3 +76,19 @@ jobs:
         run: make VERBOSE=1 check
       - name: distcheck
         run: make VERBOSE=1 distcheck
+
+  test-redhat:
+    runs-on: ubuntu-latest
+    container: registry.access.redhat.com/ubi7/ubi
```
rhansen, line 82: Is there a way to add a comment linking to a description? The hex chars in https://catalog.redhat.com/software/containers/ubi7/ubi/5c3592dcd70cc534b3a37814 make it look like it's not meant to be a permanent link.

rhansen, line 82: Why 7 instead of 8?

Can we matrix this with the CentOS equivalents?

SuperSandro2000: I wanted to test the oldest still supported os. I think there we most likely find the most incompatibilities. 

Right now we can't really combine it with centos 6 cause the checkout is different and I am sure centos 6 requires more workarounds in the future. All other use dns which is faster than yum but this is not available on redhat 7 in the base image.

rhansen: Isn't RHEL 6 still supported? I was under the impression that CentOS and RHEL are mostly equivalent, as CentOS is derived from RHEL.

SuperSandro2000: I think so but I couldn't find the Ubi for it. They introduced it only for red hat 7.

Centos 6 has an ancient version of libc which does not work with checkout@v2 and centos8 has dnf which is faster than yum. Redhat 7 should be centos7. I think it is fine if we don't share everything. When GitHub actions supports yaml anchors we can reduce the amount of yaml by a lot.

Issue #206, opened 2020-06-29
Put back ddclient file?


dkerr64:
Okay, I may have missed this when it was proposed as a PR but now that it is committed I'm running into issues.  Why have you got rid of `ddclient` file and replaced it with `ddclient.template`?  And why has the template file got rid of the `#!/usr/bin/perl` shebang.  The template file is no longer automatically recognized as a Perl file by text editors (for syntax highlighting).  And what about all existing pull requests on the original ddclient file.  I think this is going to cause major disruption for platforms that have adopted ddclient for its simplicity and install simply by copying ddclient file.

rhansen:
> Why have you got rid of `ddclient` file and replaced it with `ddclient.template`?

See 8cbcecba99324c000662badc160784c3011cfb64 and PR #161. The `ddclient` file is now "built" from the template file.

> The template file is no longer automatically recognized as a Perl file by text editors (for syntax highlighting).

Yeah, we need to add some magic strings to teach Emacs and Vim that this is a Perl file. In the meantime you'll have to manually tell your editor.

I had expected `perl ./ddclient.template` to work, but it looks like Perl itself is attempting to interpret `#!@PERL@`. I need to figure out how to disable that or work around it.

>  And what about all existing pull requests on the original ddclient file.

As [stated in the contributor guide](https://github.com/ddclient/ddclient/blob/master/CONTRIBUTING.md#git-hygiene), contributors are expected to keep their pull requests rebased on top of `master`. Git's rename detection should detect the rename and do the right thing. If `master` has diverged too much from your branch, you may need to lower the detection threshold via `-Xfind-renames=n`.

>  I think this is going to cause major disruption for platforms that have adopted ddclient for its simplicity and install simply by copying ddclient file.

It is unfortunate that users can no longer copy the script file to install it, but the long-term benefits outweigh the costs. It's not difficult to run `./configure && make`, and having a build system both makes it easier for packagers to create distribution packages and allows us to implement some important features.

dkerr64:
Just ran into another issue... the `.template `file name is interpreted by MacOS as being a Pages file (their word processor).  So it has a pages icon in finder and default open action is to open pages.  Isn't there a better way?

Same for the `.conf.template` file.  For that could we use .sample?  or just leave it as .conf?

rhansen:
[...]

PR #219, opened 2020-07-01
Update Vim modeline and Emacs file-local variables

rhansen:
Both:
  * Force the file type to Perl
  * Set the tab width to 8
  * Set the line with to 99

Emacs:
  * Disable indent-tabs-mode
  * Set the indentation level to 4

Vim:
  * Highlight column 100

Addresses #206

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -5215,7 +5215,16 @@ sub nic_cloudns_update {
 __PACKAGE__->main() unless caller() && caller() ne 'PAR';
 
 ######################################################################
-# vim: ai et ts=4 sw=4 tw=78:
+## Emacs and Vim settings
 
+# Local Variables:
+# mode: perl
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
> Set the tab width to 8

I have an 16:9 monitor but still I hate it when vim does this. Either we go with 4 or we drop this cause it is a horrible default.

rhansen:
> > Set the tab width to 8
> 
> I have an 16:9 monitor but still I hate it when vim does this. Either we go with 4 or we drop this cause it is a horrible default.

There shouldn't ever be any tabs in the file because we use spaces for indentation, so the tab width shouldn't matter. But I set it to 8 anyway because 8 is the standard width of a tab and I wanted people to cringe in horror at the sight of a tab. :smile:

PR #221, opened 2020-07-01
Add FreeBSD; change README to use json-pp for Ubuntu

thorsteneb:
Tested on FreeBSD 11.3; expected to work on 12.1 as well.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -15,6 +15,7 @@ EXTRA_DIST = \
 	sample-etc_dhclient-exit-hooks \
 	sample-etc_dhcpc_dhcpcd-eth0.exe \
 	sample-etc_ppp_ip-up.local \
+	sample-etc_rc.d_ddclient.freebsd \
 	sample-etc_rc.d_init.d_ddclient \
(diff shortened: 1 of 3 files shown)


rhansen:
Thank you!

SuperSandro2000:
There is no up to date FreeBSD Docker container so we can't add tests for it.

thorsteneb:
That is a good point, @SuperSandro2000 , and I now remember that I did need to make a change to the ddclient script as distributed: It starts with a shebang for /usr/bin/perl as well as /usr/local/bin/perl. This will cause it to fail on FreeBSD 11.3, it's necessary to remove that first line. After that, it works.

rhansen:
> It starts with a shebang for /usr/bin/perl as well as /usr/local/bin/perl. This will cause it to fail on FreeBSD 11.3, it's necessary to remove that first line. After that, it works.

This has recently changed: We now use GNU Autoconf to locate the `perl` executable, and `make` replaces the path in the shebang line, so it should just work on FreeBSD.

PR #222, opened 2020-07-01
Also run tests on Debian testing, stable, oldstable

rhansen:


.github/workflows/ci.yml
```diff
@@ -9,17 +9,28 @@ on:
       - master
 
 jobs:
-  test-ubuntu:
+  test-debian:
```
SuperSandro2000, line 12:
```suggestion
  test-debian-like:
```

I think the name fits better.

.github/workflows/ci.yml
```diff
@@ -9,17 +9,28 @@ on:
...
-    runs-on: ubuntu-${{ matrix.ubuntu }}
+        image:
+          - ubuntu:latest
+          - ubuntu:16.04
+          - debian:testing
+          - debian:stable
+          - debian:oldstable
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/library/${{ matrix.image }}
```
SuperSandro2000, line 23:
```suggestion
      image: ${{ matrix.image }}
```

This should be enough. This would also allow us to use images from different registries if we wanted.

.github/workflows/ci.yml
```diff
@@ -9,17 +9,28 @@ on:
...
     steps:
+      - name: install dependencies
+        run: |
+          apt-get update &&
```
SuperSandro2000, line 27: I am pretty sure the runner runs ``apt update`` somewhere cause the images do not contain any apt lists by default. Maybe we want to keep it just to be sure.

.github/workflows/ci.yml
```diff
@@ -9,17 +9,28 @@ on:
...
+          DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
```
SuperSandro2000, line 28: Maybe we want to set ENV via the [function for the runner](https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#about-environment-variables). Not sure.

.github/workflows/ci.yml
SuperSandro2000: Why is this comment pending?!

.github/workflows/ci.yml
rhansen: Done.

.github/workflows/ci.yml
rhansen: `apt-get update` is required. The Ubuntu VMs already have the package lists, but the Docker images don't.

.github/workflows/ci.yml
rhansen: I guess it doesn't hurt to set it everywhere, but it's only needed for this particular command so I'd like to keep in there.

.github/workflows/ci.yml
rhansen: Done.

PR #223, opened 2020-07-01
Test that the distribution tarball is complete

rhansen:


rhansen:
Set to draft until PR #222 is merged.

.github/workflows/ci.yml
```diff
@@ -39,6 +41,60 @@ jobs:
...
+          git archive -o "${tmpdir}"/git-repo.tar.gz HEAD
+          mkdir "${tmpdir}"/git-repo
+          cd "${tmpdir}"/git-repo
+          tar xvfa ../git-repo.tar.gz
+          # Delete files checked into Git that shouldn't be in the
+          # distribution tarball.
+          rm -rf \
+              .github \
+              .gitignore \
+              docs/ipv6-design-doc.md \
```
SuperSandro2000, line 65:
```suggestion
              docs/ \
```

rhansen: We might add docs here that we do want included in the tarball.

rhansen:
> We should think about putting it into some sort of CI script

Done.

.github/workflows/scripts/dist-tarball-check
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
```
SuperSandro2000, line 1: This file should have a .sh ending.

rhansen: I don't like putting suffixes on executables in case the script is rewritten in another language.

PR #224, opened 2020-07-01
Install findutils on Fedora

rhansen:


SuperSandro2000:
Why do we need this?

rhansen:
`make distcheck` uses `find`. Surprisingly it doesn't fail despite all the `find: command not found` errors.

SuperSandro2000:
> `make distcheck` uses `find`. Surprisingly it doesn't fail despite all the `find: command not found` errors.

Didn't notice that either. Can we somehow make sure this is not failing in the future? Some sort of requirement on that binary?

rhansen:
`find` is specified by POSIX, so *every* Unix-ish operating system *should* have it. The lack of `find` is almost as offensive to me as leaving out `ls`. It's so common that Autoconf itself doesn't bother checking for it. I don't think it's worth adding an check for it in our `configure.ac`.

PR #227, opened 2020-07-03
New options to specify CA certificate location

rhansen:
This is needed to run a dummy https server for `geturl()` tests.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,8 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
 
   * The `freedns` protocol (for https://freedns.afraid.org) now supports IPv6
     addresses.
+  * New `ssl_ca_dir` and `ssl_ca_file` options to specify the location of CA
+    certificates.
 
 ### Bug fixes
 
--- a/ddclient.in
+++ b/ddclient.in
@@ -383,6 +383,8 @@ my %variables = (
         'geturl'              => setv(T_STRING,0, 0, '',                   undef),
 
         'postscript'          => setv(T_POSTS, 0, 0, '',                   undef),
+        'ssl_ca_dir'          => setv(T_FILE,  0, 0, undef,                undef),
+        'ssl_ca_file'         => setv(T_FILE,  0, 0, undef,                undef),
     },
     'service-common-defaults' => {
         'server'              => setv(T_FQDNP, 1, 0, 'members.dyndns.org', undef),
@@ -760,6 +762,8 @@ my @opt = (
     [ "options",     "=s",  "-options opt,opt     : optional per-service arguments (see below)" ],
     "",
     [ "ssl",         "!",  "-{no}ssl              : do updates over encrypted SSL connection" ],
+    [ "ssl_ca_dir",  "=s", "-ssl_ca_dir           : directory containing certificates of trusted certificate authorities (default: auto-detect)" ],
+    [ "ssl_ca_file", "=s", "-ssl_ca_file          : file containing certificates of trusted certificate authorities (default: auto-detect)" ],
     [ "retry",       "!",  "-{no}retry            : retry failed updates." ],
     [ "force",       "!",  "-{no}force            : force an update even if the update may be unnecessary" ],
     [ "timeout",     "=i", "-timeout max          : wait at most 'max' seconds for the host to respond" ],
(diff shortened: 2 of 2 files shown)


SuperSandro2000:
Did you forget to add tests or is this just a note for the future=

rhansen:
Tests will come in a future PR. This is just a prerequisite to adding those tests.

dkerr64:
Or we can just add support for `curl` and let it handle it.  I really think that the intricacies of network requests should be left to something else, like `curl`...
```
    if (opt('curl')) {
        if (!`/usr/bin/curl --version 2>/dev/null`) {
            fatal("You requested network access with curl but curl is not installed.");
        }
        # Access network using curl.
        if (!opt('exec')) {
            debug("skipped network connection");
            verbose("SENDING:", "%s", "${server}/${url}");
        } elsif (($headers ne '')) {
            my $timeout    = opt('timeout');
            my @head_lines = split('\n', $headers);
            my $head_args  = '';
            foreach my $head_line (@head_lines) {
                $head_args .= "-H '$head_line' ";
            }

            # don't include ${url} as that might expose login credentials
            $0 = sprintf("%s - curl (w/headers) sending to %s", $program, "https://${server}");
            verbose("SENDING:", "curl (w/headers) to %s", "https://${server}");
            $reply = <<`USE_CURL_WITH_HEADERS`;
/usr/bin/curl -s -X $method --user-agent '${program}/${version}' \\
              --connect-timeout $timeout --max-time $timeout \\
              ${head_args} \\
              --data '${data}' \\
              --url 'https://${server}/${url}' 2>/dev/null
USE_CURL_WITH_HEADERS

            if (!$reply) {
                # don't include ${url} as that might expose login credentials
                warning("curl (w/headers) cannot connect to https://%s using IPv%s",${server},$ipversion);
            }
        } else {
            my $timeout  = opt('timeout');
            my $auth     = '';
            my $curlopt  = '';
            my $protocol = 'http';
            $auth    = "--user '${login}:${password}'" if (($login ne '') && ($password ne ''));
            $curlopt = '--ipv4'                        if ($ipversion eq '4');
            $curlopt = '--ipv6'                        if ($ipversion eq '6');
            if ($use_ssl) {
                $protocol = 'https';
                $curlopt .= ' --insecure';
            }

            # don't include ${url} as that might expose login credentials
            $0 = sprintf("%s - curl sending to %s", $program, "${protocol}://${server}");
            verbose("SENDING:", "curl to %s", "${protocol}://${server}");
            $reply = <<`USE_CURL`;
/usr/bin/curl -s ${auth} --user-agent '${program}/${version}' \\
              --connect-timeout $timeout --max-time $timeout $curlopt \\
              --url '${protocol}://${server}/${url}' 2>/dev/null
USE_CURL
            if (!$reply) {
                # don't include ${url} as that might expose login credentials
                warning("curl cannot connect to %s://%s using IPv%s",${protocol},${server},$ipversion);
            }
        }    
    } else {
        # Access network using perl functions. 
    }
```

SuperSandro2000:
Curl is definitely a separate big task to do and we need good test coverage or we break everything along to way to 4.0.0.

dkerr64:
> Curl is definitely a separate big task to do and we need good test coverage or we break everything along to way to 4.0.0.

I disagree.  Adding support for curl is easy, the entire code is in above comment, and it is WAY more reliable approach than using Perl modules because all the networking stuff is handled for you. Using Perl modules is in fact the bigger task and fraught with risk as demonstrated by this very pull request and the test suite for using the Perl modules (if it existed) would need to be much more thorough that that for curl.

Curl has been used exclusively by the [ddclient-curl](https://github.com/astlinux-project/ddclient-curl) fork for over 4 years now without a problem... and indeed the whole reason that fork was created was because of all the issues trying to use Perl modules for networking and that @wimpunk appeared to have abandoned it.

Now that ddclient is being actively worked on again curl support should be added sooner, not later, so that it can be adopted on platforms that do not have the Perl modules required for SSL and IPv6.

@abelbeck

rhansen:
This is not the place to discuss the merits of `curl` and the strategy to switch to it. Please file a new issue.

abelbeck:
> This is not the place to discuss the merits of curl and the strategy to switch to it.

@rhansen : But it is related, as this PR is not needed at all if `curl` is used exclusively for network IO.

rhansen:
Whether `geturl` uses `curl` or something else doesn't matter—users and unit tests will always need to be able to tell ddclient where CA certificates can be found.

PR #234, opened 2020-07-04
Fix URL parsing for IPv6 URLs and pathless URLs

rhansen:
Now the following valid URLs are parsed correctly:
  * `http://[::1]:123/foo`
  * `http://localhost?foo=bar`

There are still problems with the URL parsing logic but this is enough to write some unit tests.


SuperSandro2000:
Re running the tests. Not sure why they failed.

Issue #232, opened 2020-07-03
Add support for curl as an option in geturl()


dkerr64:
ddclient currently uses Perl modules for network access.  As support for SSL and IPv6 is added this places additional requirements on host systems to have the necessary modules that are not always installed by default or even available.  An alternate method for network access would be to use curl.

This off loads the intricacies to another package and gives the user choice based on their requirements or system capabilities.  This would be an additional option and not remove current network code (at least initially)

rhansen:
PR #233 runs the curl binary, but I'm not convinced that is the appropriate way to go. [WWW::Curl](https://metacpan.org/pod/WWW::Curl) is available even on old distributions, and using it might avoid hassles such as escaping metacharacters or keeping error codes in sync.

[LWP](https://metacpan.org/pod/LWP) is also ubiquitous, and integrates well with other Perl modules such as HTTP::Request. I believe it is also pure Perl (no C extensions).

dkerr64:
> PR #233 runs the curl binary, but I'm not convinced that is the appropriate way to go. [WWW::Curl](https://metacpan.org/pod/WWW::Curl) is available even on old distributions, and using it might avoid hassles such as escaping metacharacters or keeping error codes in sync.
> 
> [LWP](https://metacpan.org/pod/LWP) is also ubiquitous, and integrates well with other Perl modules such as HTTP::Request. I believe it is also pure Perl (no C extensions).

WWW::Curl is not installed by default on any system I tried.

The primary issue is maintaining compatibility with the simplest of systems, and specifically router/gateway devices that might be running things like DD-WRT or similar.  These systems cannot easily add Perl modules but they do typically have curl (or wget) as cmd-line tools.  So my goal is not to replace IO::System with something more elegant, it is to provide something that will work on the most basic of systems. @abelbeck FYI.

SuperSandro2000:
> even on old distributions

Centos 6 has a package for it and it usually has no package for anything we want to use.

https://centos.pkgs.org/6/centos-x86_64/perl-WWW-Curl-4.09-4.el6.x86_64.rpm.html

> WWW::Curl is not installed by default on any system I tried.

If you are going to install ddclient with your package manager this is not a problem.

> DD-WRT

Has a [package system](https://wiki.dd-wrt.com/wiki/index.php/Quick_list_of_Optware_packages) and then the devs need to add one package for ddclient. I think this shouldn't be to much extra work.

> Perl modules but they do typically have curl (or wget) as cmd-line tools

[...]

PR #235, opened 2020-07-04
Add `geturl` connectivity unit tests

rhansen:
Addresses #232


rhansen:
Looks like the GitHub runner's don't support IPv6:  https://github.com/actions/virtual-environments/issues/668

.github/workflows/ci.yml
```diff
@@ -29,7 +29,14 @@ jobs:
               automake \
               ca-certificates \
               git \
+              libhttp-daemon-perl \
+              libhttp-daemon-ssl-perl \
```
SuperSandro2000, line 33: ``libhttp-daemon-ssl-perl`` requires ``libhttp-daemon-perl`` already so the later can be removed.

SuperSandro2000:
> Looks like the GitHub runners don't support IPv6 and there's no plan to add support: [actions/virtual-environments#668](https://github.com/actions/virtual-environments/issues/668)

If we only need localhost I think we can monkey patch it in like the person here did https://github.com/actions/virtual-environments/issues/929#issuecomment-648209795

configure.ac
```diff
@@ -53,7 +53,17 @@ m4_foreach_w([_m], [
...
+    Exporter
+    HTTP::Daemon=6.12
+    HTTP::Daemon::SSL
+    HTTP::Message::PSGI
+    IO::Socket::IP
+    Scalar::Util
+    Test::TCP
     Test::Warnings
+    Time::HiRes
+    URI
```
SuperSandro2000, line 66: Can we move this to error? This makes getting the other tests complete easier.

rhansen: Modules used only by tests shouldn't cause `configure` to fail because the user isn't required to run `make check` to build and install ddclient.

rhansen:
> If we only need localhost I think we can monkey patch it in like the person here did [actions/virtual-environments#929 (comment)](https://github.com/actions/virtual-environments/issues/929#issuecomment-648209795)

Unfortunately that won't help us. We need to listen on ::1, which requires IPv6 support to be enabled in the runner's kernel.

SuperSandro2000:
Not sure what to do about this issue. I am going to add a commit later to install the modules in centos, fedora, etc.

rhansen:
There's nothing that can be done other than wait for GitHub to add IPv6 support. Docker containers aren't full virtual machines so insmod won't work (the kernel is shared between the host and the container).

The IPv6 tests are skipped if the runner doesn't support IPv6, so nothing will break. We'll just have to run the tests manually to exercise the IPv6 code.

configure.ac
SuperSandro2000: Yeah, you are right. Missed that.

SuperSandro2000:
I think I give up on centos{6,8} and redhat in the CI. They are basically missing everything and you need to use cpan to install all the modules. There is not even cpanm in the package repo....

SuperSandro2000:
I don't know if it is necessary but I could provide a VM with a selfhosted runner which has ipv6 enabled but for now we probably can just manually execute the pipeline for the other dists.

SuperSandro2000:
We should merge this nevertheless if it is ready no matter if actions support ipv6 or not.

rhansen:
> I think I give up on centos{6,8} and redhat in the CI. They are basically missing everything and you need to use cpan to install all the modules. There is not even cpanm in the package repo....

It's OK to not install test-only modules—the tests that depend on those modules will just be skipped. It would be nice to run those tests, but it's not worth the effort to install the modules.

> 
> I don't know if it is necessary but I could provide a VM with a selfhosted runner which has ipv6 enabled but for now we probably can just manually execute the pipeline for the other dists.

I think we can do without for now.

> 
> We should merge this nevertheless if it is ready no matter if actions support ipv6 or not.

It's ready, except PR #234 must be merged first.

PR #238, opened 2020-07-05
Use `subtest` to group related tests

rhansen:
This reduces the output of `make check`, making it easier to zero in on results of interest.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -76,7 +76,73 @@ generated_tests = \
 	t/version.pl
 TESTS = $(handwritten_tests) $(generated_tests)
 EXTRA_DIST += $(handwritten_tests) \
+	t/lib/Test/Builder.pm \
+	t/lib/Test/Builder/Formatter.pm \
+	t/lib/Test/Builder/IO/Scalar.pm \
+	t/lib/Test/Builder/Module.pm \
+	t/lib/Test/Builder/Tester.pm \
+	t/lib/Test/Builder/Tester/Color.pm \
+	t/lib/Test/Builder/TodoDiag.pm \
+	t/lib/Test/More.pm \
+	t/lib/Test/Simple.pm \
+	t/lib/Test/Tester.pm \
+	t/lib/Test/Tester/Capture.pm \
+	t/lib/Test/Tester/CaptureRunner.pm \
+	t/lib/Test/Tester/Delegate.pm \
+	t/lib/Test/use/ok.pm \
+	t/lib/Test2.pm \
+	t/lib/Test2/API.pm \
+	t/lib/Test2/API/Breakage.pm \
+	t/lib/Test2/API/Context.pm \
+	t/lib/Test2/API/Instance.pm \
+	t/lib/Test2/API/Stack.pm \
+	t/lib/Test2/Event.pm \
+	t/lib/Test2/Event/Bail.pm \
+	t/lib/Test2/Event/Diag.pm \
+	t/lib/Test2/Event/Encoding.pm \
+	t/lib/Test2/Event/Exception.pm \
+	t/lib/Test2/Event/Fail.pm \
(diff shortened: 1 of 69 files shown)


SuperSandro2000:
Centos 6...
```
FAIL: t/parse_assignments.pl (exit: 255)
========================================

Useless use of hash element in void context at ./t/parse_assignments.pl line 55.
Useless use of reference constructor in void context at ./t/parse_assignments.pl line 55.
ok 1 # skip Can't locate Test/Warnings.pm in @INC (@INC contains: /__w/ddclient/ddclient /__w/ddclient/ddclient/t/lib /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at ./t/parse_assignments.pl line 4.
# 
Can't call method "subtest" on unblessed reference at ./t/parse_assignments.pl line 55.
# Tests were run but no plan was declared and done_testing() was not seen.
```

rhansen:
Sigh. Looks like the version of Test::More that comes with Perl v5.10.1 doesn't have subtest.

SuperSandro2000:
I am not sure how we want to handle test modules for centos 6, 8 and redhat because they are either out of date or missing completely. Do we want to use Cpanminus to install them all?

rhansen:
I'm concerned that using cpanminus to install modules on CentOS/RHEL might mask compatibility problems.

For example, suppose ddclient itself (not tests) adds a dependency on module Foo. There is a Foo package for CentOS 6, but it's a really old version. That's OK—we're only using features in Foo that exist in the old version. Later, we write a test that depends on module Bar. There is no Bar package for CentOS 6, so we use cpanminus to install Bar. Unknown to us, cpanminus brings in a new version of Foo because Bar depends on the newer version of Foo. Later, we modify ddclient to use a feature that is only available in newer versions of Foo. All of the CentOS 6 tests still pass (because the new version of Foo is installed during tests) even though we are using a feature that is not available in the CentOS 6 Foo package.

If there isn't an official package for a particular module, or if the version isn't new enough, I think our safest options are:
  * Skip the tests that depend on (the newer version of) the module.
  * Copy the module into our `t/lib` directory. (We shouldn't copy any modules that are used by ddclient itself because it might mask compatibility problems.)

rhansen:
I copied the latest [Test::Simple](https://metacpan.org/release/Test-Simple) package to `t/lib` so that we can use `subtest`. I don't like to do that, but I think it's better than the alternatives. Once we drop support for CentOS 6 we can revert the commit to delete the files.

SuperSandro2000:
Can we add a submodule instead of copying the files? The GitHub repo is located here https://github.com/Test-More/test-more/ and we can just track the tag we need.

rhansen:
Submodules add considerable complexity with negligible benefit:
  * Users can't simply `git clone`—they will also have to `git submodule update --init`. This is a significant barrier for users who aren't familiar with Git.
  * It's unclear to me whether we have to `git submodule update --init` in the CI runners or if the checkout action will do that for us.
  * It's unclear to me whether `git archive` understands submodules. If not, we will have to update the distribution tarball check.
  * We wouldn't be able to put the Perl modules inside `t/lib`, which means we would have to add another `-I` flag specifically for that submodule.
  * We can't easily make local changes if needed.

It is almost always better to do a subtree merge than to use submodules, but even that is more complicated than we need.

SuperSandro2000:
> Users can't simply git clone

You can git clone --recurse-submodules

We can figure out the ci but git archive does not work with submodules by default.

> We wouldn't be able to put the Perl modules inside t/lib, which means we would have to add another -I flag specifically for that submodule.

I did not check that. Bummer.

I do not want I subtree merge and with changing paths it is even harder. The current way is probably the easiest and only needed for legacy distros.

PR #241, opened 2020-07-06
Fix detection of `get_ip` caller

rhansen:
This was broken in f6f920eb3971a0d239b564fb53d7176e97ece9be.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -28,6 +28,7 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
   * Fixed a regression introduced in v3.9.0 that caused
     `use=ip,ip=<ipv4-address>` to fail.
   * "true" is now accepted as a boolean value.
+  * The `ssl` option now applies to the `web` URL.
 
 ### Compatibility and dependency changes
 
--- a/ddclient.in
+++ b/ddclient.in
@@ -2166,11 +2166,7 @@ sub get_ip {
         $arg = $url;
 
         if ($url) {
-            $reply = geturl({
-                proxy => opt('proxy', $h),
-                url => $url,
-                ignore_ssl_option => 1,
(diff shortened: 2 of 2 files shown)


dkerr64:
That section of code is quite cryptic.  I might suggest the following cleanup...
```
debug("opt(fw = %s)", opt('fw')) if opt('fw');
debug("glo fw = %s", $globals{'fw'}) if $globals{'fw'};
## always omit SSL for connections to local router
my $is_get_ip = (((caller(1))[3] // '') eq 'ddclient::get_ip');
if ($force_ssl || ($globals{'ssl'} && !$is_get_ip)) {
   ## etc
}
```
The reason to create a $`is_get_ip` is because we will need to also test for `ddclient::get_ipv4` and `ddclient::get_ipv6` once that code is merged (I have it ready and waiting... pending my other PRs getting merged first).  So it will just make it much easier to understand... and I also want to use `$is_get_ip` later in my curl implementation, so I need the test twice.

dkerr64:
Oh... and doing `debug() if xyz` is so much clearer than `xyz && debug()`. At least to my eyes.

rhansen:
> That section of code is quite cryptic. I might suggest the following cleanup...

I took a different approach; please take another look.

> Oh... and doing `debug() if xyz` is so much clearer than `xyz && debug()`.

Agreed, but out of scope for this PR.

dkerr64:
> > That section of code is quite cryptic. I might suggest the following cleanup...
> 
> I took a different approach; please take another look.

Yes, your approach is better.

> 
> > Oh... and doing `debug() if xyz` is so much clearer than `xyz && debug()`.
> 
> Agreed, but out of scope for this PR.

And again you found a better answer but just deleting them.

rhansen:
Thanks for the review!

PR #248, opened 2020-07-07
Add autoconf checks for all used core modules

rhansen:
Also: Import the `AF_INET` and friends constants from Socket.


rhansen:
OK, this is weird. Why does adding autoconf checks cause `make check` to fail?

rhansen:
Looks like I lost a race condition in GitHub itself. To fix it I ran `git commit --amend --reset-author -C HEAD` to generate a different commit object and force-pushed the result.

PR #249, opened 2020-07-08
Fix `write_cache` tests

rhansen:
* Enable autoflush on `STDOUT` and `STDERR` for tests
  * Don't match against locale-dependent `$!`
  * Don't test writing to a read-only directory
  * Switch from `redefine` to `mock`
  * Add missing test dependency on Test::MockModule. (This should have been done in 6ae5fe62d7a5d60c4a875258ea422d59a744cacc but I forgot.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
               libio-socket-inet6-perl \
               libio-socket-ip-perl \
               libplack-perl \
+              libtest-mockmodule-perl \
               libtest-tcp-perl \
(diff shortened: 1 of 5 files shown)


SuperSandro2000:
The pipeline needs fixing. Can you do that?

rhansen:
Yeah, I'm currently trying to figure out what went wrong. I wonder if it's because the processes run as root and root doesn't get any permission denied errors when writing to a read-only directory.

rhansen:
OK, it should be working now.

PR #250, opened 2020-07-09
Adjust path to ddclient

Steve8291:
Installing with `make install` automatically places the ddclient executable in "/usr/bin/ddclient"
This service file has the wrong path and that causes systemd to be unable to start ddclient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ In my case, it is named dhcpcd-eth0.exe and contains the lines:
 
 ```shell
 #!/bin/sh
-PATH=/usr/sbin:/root/bin:${PATH}
+PATH=/usr/bin:/root/bin:${PATH}
 logger -t dhcpcd IP address changed to $1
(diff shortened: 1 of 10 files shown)


rhansen:
The sample files currently aren't installed anywhere, so there's not much point in substituting `@bindir@`. I think for now we should just replace all instances of `/usr/sbin` with `/usr/bin`. In the future we may want to write an Autoconf test to determine the init system so that we can install the appropriate init script in `/etc/init.d` or `/etc/systemd/system` or wherever. At that point we can substitute `@bindir@`.

@Steve8291: Would you mind updating your pull request to replace all instances of `/usr/sbin` with `/usr/bin`?

Steve8291:
@rhansen I think I got all of them. Someone should definitely double check what I did. I'm assuming that in other distros the install location is `/usr/bin` as well.

rhansen:
Thank you!

PR #256, opened 2020-07-12
Import IO::Socket::INET

### LINKED ISSUE #255: Error message... Can't locate object method "new" via package "IO::Socket::INET"
I got the following error running `ddclient.in - query` from master branch...
```
Can't locate object method "new" via package "IO::Socket::INET" (perhaps you forgot to load "IO::Socket::INET"?) at ./ddclient.in line 2106.
```
If I add `use IO::Socket::INET` immediately before the call then it works.

rhansen:
We've always required IO::Socket::INET but never explicitly included it. For some reason that hasn't been a problem until now.

Fixes #255

cc @dkerr64

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,6 +40,7 @@ m4_foreach_w([_m], [
     File::Basename
     File::Path
     Getopt::Long
+    IO::Socket::INET
     Socket
     Sys::Hostname
     version=0.77
--- a/ddclient.in
+++ b/ddclient.in
@@ -25,6 +25,7 @@ use warnings;
 use File::Basename;
 use File::Path qw(make_path);
 use Getopt::Long;
+use IO::Socket::INET;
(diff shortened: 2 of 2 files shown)


dkerr64:
At the top of the file there was always a `use IO::Socket` which was sufficient.  At some point that changed to `use Socket qw(AF_INET AF_INET6 PF_INET PF_INET6);`...  note the missing `IO::` if you add in the `IO::` to that line then it works again.  Are there even modules for socket without the IO:: ???  I think the missing IO:: is the bug.

rhansen:
>  I think the missing IO:: is the bug.

No, `Socket` without `IO::` is intentional. There is a Socket module: https://perldoc.perl.org/Socket.html

Maybe `use IO::Socket;` pulls in IO::Socket::INET which is why it worked before but doesn't now.

PR #258, opened 2020-07-15
Fix `-daemon` command-line argument parsing

rhansen:
It turns out the logic removed in d10601de11d287cf8a44a043ed31d658cf219a51 had a purpose after all.


rhansen:
I'm going to merge this without review because the current daily PPA package is updating every 5 seconds instead of every 5 minutes, which is inevitably going to get users banned. I have tested it on my machine.

PR #265, opened 2020-08-02
Add additional builtinweb entries

rhansen:
Note that this PR also renames `ipifyipv4` and `ipifyipv6` to `ipify-ipv4` and `ipify-ipv6` to make them easier to read.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -22,6 +22,7 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
     certificates.
   * New built-in IP discovery service shorthands:
       - `googledomains` from https://domains.google
+      - `he` from https://he.net
       - `ipify-ipv4` and `ipify-ipv6` from https://www.ipify.org
       - `myonlineportal` from https://myonlineportal.net
       - `noip-ipv4` and `noip-ipv6` from https://www.noip.com
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -20,6 +20,9 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
     addresses.
   * New `ssl_ca_dir` and `ssl_ca_file` options to specify the location of CA
     certificates.
+  * New built-in IP discovery service shorthands:
+      - `ipifyipv4` and `ipifyipv6` from https://www.ipify.org
+      - `myonlineportal` from https://myonlineportal.net
   * New built-in shorthands for obtaining the IP address from the following
     devices ([thanks to Geoff Simmons](https://bugs.debian.org/589980)):
       - `alcatel-530`: Alcatel/Thomson SpeedTouch 530
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -21,6 +21,7 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
   * New `ssl_ca_dir` and `ssl_ca_file` options to specify the location of CA
     certificates.
   * New built-in IP discovery service shorthands:
+      - `googledomains` from https://domains.google
       - `ipify-ipv4` and `ipify-ipv6` from https://www.ipify.org
       - `myonlineportal` from https://myonlineportal.net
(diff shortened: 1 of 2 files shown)


dkerr64:
While testing my IPv6 code (using `ddclient -query`) I have updated the `test_possible_ip` function to call the builtinweb entries three ways... forcing IPv4, forcing IPv6 and letting the system decide.  When forcing IP version some of the web sites fail because they do not accept access by that method.

So... should we update this list of servers to include identifying which accept IPv4 and which accept IPv6.  Or do we just let ddclient post the error when it fails?

SuperSandro2000:
> So... should we update this list of servers to include identifying which accept IPv4 and which accept IPv6. Or do we just let ddclient post the error when it fails?

We should probably only do this for domains which are named after ipv4 or ipv6 because otherwise we have an endless task. 

Or we just use let the system decide or use DNS entries.

dkerr64:
I would also propose adding these to the list...
```
     'he'             => { 'url' => 'http://checkip.dns.he.net/', 'skip' => 'Your IP address is', }, 
     'whatismyv6'     => { 'url' => 'http://whatismyv6.com/', 'skip' => 'Address of:', },
     'nsupdate.info'  => { 'url' => 'https://ipv6.nsupdate.info/myip', , },
```

David

dkerr64:
> We should probably only do this for domains which are named after ipv4 or ipv6 because otherwise we have an endless task.
> 
> Or we just use let the system decide or use DNS entries.

I'm content to leave it as is for now... let the error occur.

Also, FYI, when querying web sites that support both IPv4 and IPv6 there is a bias towards returning the IPv6 address.  My results when not forcing IPv4 or IPv6...
```
use=web, web=dyndns address is xxx.yyy.157.237
use=web, web=freedns address is xxx.yyy.157.237
use=web, web=googledomains address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=he address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=ipify-ipv4 address is xxx.yyy.157.237
use=web, web=ipify-ipv6 address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=loopia address is xxx.yyy.157.237
use=web, web=myonlineportal address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=noip-ipv4 address is xxx.yyy.157.237
use=web, web=noip-ipv6 address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=nsupdate.info address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=whatismyv6 address is 2601:aaaa:bbbb:e723:bdff:db5a:dfbc:d3e1
use=web, web=zoneedit address is xxx.yyy.157.237
```
Which is one reason why we need separate `get_ipv4` and `get_ipv6` functions.

rhansen:
> So... should we update this list of servers to include identifying which accept IPv4 and which accept IPv6. Or do we just let ddclient post the error when it fails?

I debated this, but decided not to add `-ipv4` suffixes because the services don't say anything about IPv6 in their documentation and they might enable IPv6 in the future. The services are ambiguous about IPv6 support, so we don't have much of a choice other than to be ambiguous ourselves.

> I would also propose adding these to the list...

Thanks for those, I'll add them.

> when querying web sites that support both IPv4 and IPv6 there is a bias towards returning the IPv6 address.

That's expected. If there is a AAAA record, IO::Socket::SSL will use IPv6 if supported by the system.

> Which is one reason why we need separate `get_ipv4` and `get_ipv6` functions.

Yup!

rhansen:
> > I would also propose adding these to the list...
> 
> Thanks for those, I'll add them.

Done.

dkerr64:
> > > I would also propose adding these to the list...
> > 
> > 
> > Thanks for those, I'll add them.
> 
> Done.

Pretty comprehensive list now !

PR #266, opened 2020-08-03
Add get_ip_from_interface function

dkerr64:
This adds `get_ip_from_interface` with support for both IPv4 and IPv6 and updates `get_ip` to call this rather than call system commands itself.  Test cases are includes for many flavors of Linux... Ubuntu, RHEL, Busybox, Buildroot, MacOS, FreeBSD.  This also accepts `default` as an interface name thus satisfying the request in PR #50

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
               libtest-tcp-perl \
               libtest-warnings-perl \
               liburi-perl \
+              net-tools \
               make \
               ;
       - uses: actions/checkout@v2
@@ -117,6 +118,7 @@ jobs:
               perl-Test-MockModule \
               perl-Test-TCP \
               perl-Test-Warnings \
+              net-tools \
               ;
       - name: autogen
         run: ./autogen
@@ -142,6 +144,7 @@ jobs:
               perl-HTTP-Daemon \
               perl-IO-Socket-INET6 \
               perl-core \
+              iproute \
               ;
       - name: autogen
         run: ./autogen
--- a/Makefile.am
+++ b/Makefile.am
@@ -71,6 +71,7 @@ AM_PL_LOG_FLAGS = -Mstrict -w \
 	-I'$(abs_top_srcdir)'/t/lib \
 	-MDevel::Autoflush
 handwritten_tests = \
+	t/get_ip_from_if.pl \
 	t/geturl_ssl.pl \
 	t/is-and-extract-ipv4.pl \
 	t/is-and-extract-ipv6.pl \
--- a/ddclient.in
+++ b/ddclient.in
@@ -2706,11 +2706,68 @@ my $regex_ipv6_ula = qr{
 }xi;
 
 ######################################################################
+## get_default_interface finds the default network interface based on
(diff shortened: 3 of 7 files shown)


dkerr64:
Adding tests to check retrieval of IPv4 and IPv6 addresses from the test systems results in failure for all except CentOS 8.  It appears others do not have `ip` or `netstat` commands (and maybe not `ifconfig`).  Can those be added to the test systems... actually some should have `ip`, some not and those that don't should have `netstat` and `ifconfig` instead.

Thanks!

SuperSandro2000:
You can install packages for the tests in https://github.com/dkerr64/ddclient/blob/get-if-ip/.github/workflows/ci.yml.

Alls variants of Debian and Ubuntu should support ``iproute2`` which would mean we do not test ``netstat`` at all. I am not sure how to handle this correctly.  For Centos, etc I always use [pkgs.org](https://pkgs.org/search/?q=iproute) to find the right packages.

@rhansen What do you think about this? How do we want to handle ip/netstat. Maybe test both in one test?

rhansen:
I find it's much easier and faster to review many small PRs vs. one large PR. Would you please move the `get_default_interface` feature to a follow-up PR? That way we won't hold up the bulk of this PR as we figure out the best way to test `ip route` and `netstat`.

dkerr64:
> I find it's much easier and faster to review many small PRs vs. one large PR. Would you please move the `get_default_interface` feature to a follow-up PR? That way we won't hold up the bulk of this PR as we figure out the best way to test `ip route` and `netstat`.

I can do that.  But in the meantime I'm going to turn on debugging so I can see more of what is actually going on.

dkerr64:
So the CI tests were really helpful.  I found that I had to make sure not to return the loopback interface, and that I didn't return a default route for IPv4 when asked for IPv6.  That the CI systems don't all have IPv6 nor the required net-tools or iproute was really helpful and prompted me to greatly improve my debug messages.

I have added net-tools to some of the CI systems and iproute to RHEL.  The iproute (or iproute2 on debian/ubuntu) is preferred and consistent across platforms, so I only installed it on one.  The net-tools package is more variable and is fallback, so I wanted to have it on most (without iproute, so ddclient uses them).  We may want to change this in the future.  But in the meantime things are good.

dkerr64:
@rhansen I looked at how to split this into two PRs but discovered that it is pretty ugly... github doesn't really have a way to make one PR dependent on another PR and allow for easy merging/rebase. 

So what I did was rebase and squash into a two separate commits so you can see the changes related to adding `get_ip_from_interface` (and the test case updates) in the first commit, then the changes related to adding `get_default_interface` in the second commit.

Then I found a missing comment block which should probably be its own PR, but well I am lazy today so put it here for visibility.

dkerr64:
@rhansen @SuperSandro2000 gentle nudge to ask for review and merge.  This is needed before I can offer up any more pull requests.  Thanks.

SuperSandro2000:
@dkerr64 Can you resolve the merge conflict? After that I want to merge this.

dkerr64:
> @dkerr64 Can you resolve the merge conflict? After that I want to merge this.

@SuperSandro2000 Done.

PR #268, opened 2020-08-03
New command-line options to list protocols, IP discovery services, and devices

rhansen:
* Move `use=cisco` and `use=cisco-asa` to `%builtinfw`
  * New `-list-web-services` option
  * New `-list-protocols` option
  * New `-list-devices` option

Motivation: Make it possible for the Debian package to present a list of choices to the user during installation.


rhansen:
I'm not a huge fan of the name `-list-web-services`, but I couldn't think of something better. Can you?

SuperSandro2000:
I cherry-picked all list commits. I don't think we need to change anything about cisco as it is already deprecated.

PR #269, opened 2020-08-03
Turn `if-skip` into a no-op and mark it as deprecated

rhansen:
There is no way the user can meaningfully set `if-skip` because the user doesn't have control over how ddclient reads an interface's settings (ddclient could theoretically run `ip addr show`, run `ifconfig`, read a file in `/dev`, make a system call, use a Perl library, etc.).

cc @dkerr64


dkerr64:
Two thumbs up from me.  Thanks.

PR #274, opened 2020-08-09
Redact login and password when printing out internal hash values

dkerr64:
When running with debug and verbose the login and password for a service may be printed out.  This redacts that from the printout.


SuperSandro2000:
I think this is quite useful to know if the right credentials are used. Not sure how we want to handle debug and verbose for bug reports.

dkerr64:
People tend to copy/paste debug output to public places when they have a problem and seek help.  Not everyone is careful enough to remove private information.  It is best practice to remove all credentials from debug output.

SuperSandro2000:
Debug and verbose is a mess anyway and very confusing. Maybe we should redesign it and while we are doing this add a flag which shows credentials but is totally different from the current flags.

dkerr64:
Right now there is `verbose`, `info`, `debug` and `debug2`.  Verbose and info both display with `-verbose` flag so there is no difference (other than printing INFO at the front of the line).  Debug2 displays if both `-debug` and `-verbose` are set -- but it is not used once in the code, which is probably a good thing as we could then change that.

SuperSandro2000:
I think we should remove ``info`` and ``debug2``. ``verbose`` and ``debug`` are already confusing enough.

SuperSandro2000:
This does not work for providers like cloudflare which identify via a token. Do you think this can be easily added?

Issue #172, opened 2020-06-12
Comprehensive IPv6 support


rhansen:
This issue is to help us track everything involved in adding full IPv6 support:

  * [ ] Create a design document: #163
  * [ ] Add the ability to force IPv4 or IPv6 in `geturl`: #169
  * [ ] Implement the new configuration settings mentioned in the design document
  * [ ] Deprecate the legacy configuration settings
  * [ ] Protocol support:
      * [ ] inwx.com: #96
      * [ ] DynDNS: #75
      * [ ] nsupdate
      * [ ] others...
  * [ ] Remove deprecated `'wantip'` once all protocols have been updated to `'wantipv4'`/`'wantipv6'`.
  * [ ] Ideally `geturl` would implement [RFC 8305](https://tools.ietf.org/html/rfc8305). Probably the easiest way to do this is to switch to curl.

Greg81:
Having support for no-uip.com would be great :)

dkerr64:
> Having support for no-uip.com would be great :)

Cannot find no-uip.com.  Did you mean https://noip.com ?

Greg81:
Yes, I am sorry for my fat fingers and lack of error checking.

dkerr64:
PR #291 opened with the framework for supporting IPv6

beadon:
This is something I could sink my teeth into .  IS this still an ongoing issue with the 4.x release ?

PR #291, opened 2020-09-23
IPv6 framework + Cloudflare + FreeDNS

dkerr64:
There is no way to add IPv6 support in piece parts so here goes...

Following guidance in the design document #163 and issue #172 ...

1. Added all the options and data structures needed, marking those old ones as deprecated. 
2. Added new `get_ipv4()` and `get_ipv6()` functions
3. Updated `test_possible_ips()` function to add IPv4 and IPv6
4. Updated `run_postscript()` function to pass both IPv4 and IPv6
5. Updates to `write_cache()` and `init_config()` to support IPv4 and IPv6
6. Updated `update_nics()` and `nic_updatable()` functions for IPv4 and IPv6
7. Added `max-warn` option to prevent flooding syslog with warnings because we may be only updating one of IPv6 or IPv4 and not both.

And...
Updated **Cloudflare** and **FreeDNS** functions to work with the new framework.  All other services will continue to work with the old and can be updated over time, but I needed to include one or two in this PR so it can be tested.

Note that config option `ipv6` changes from `T_BOOL` to `T_IPV6`.  This is only (that I am aware of) potential conflict with existing usage.

Requires **review and testing** by more than just me.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -75,6 +75,9 @@ my ($result, %config, %cache);
 my $saved_cache;
 my %saved_opt;
 my $daemon;
+# Control how many times warning message logged for invalid IP addresses
+my (%warned_ip, %warned_ipv4, %warned_ipv6);
+my $inv_ip_warn_count = opt('max-warn') // 1;
 
 sub T_ANY    { 'any' }
 sub T_STRING { 'string' }
@@ -90,9 +93,13 @@ sub T_FILE   { 'file name' }
 sub T_FQDNP  { 'fully qualified host name and optional port number' }
 sub T_PROTO  { 'protocol' }
 sub T_USE    { 'ip strategy' }
+sub T_USEV4  { 'ipv4 strategy' }
+sub T_USEV6  { 'ipv6 strategy' }
 sub T_IF     { 'interface' }
 sub T_PROG   { 'program name' }
 sub T_IP     { 'ip' }
+sub T_IPV4   { 'ipv4' }
+sub T_IPV6   { 'ipv6' }
 sub T_POSTS  { 'postscript' }
 
 ## strategies for obtaining an ip address.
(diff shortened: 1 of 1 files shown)


DivineDominion:
Some feedback: I get no errors when I use the `usev6=ip,...` setting, but it also doesn't do anything with dynv6.net (dyndns2 protocol), not even output errors or warnings in `-debug -verbose -noquiet` mode, which is strange.

dkerr64:
@DivineDominion right... for this PR I just implemented Cloudflare and FreeDNS.  Once this is merged I can work on the other services, but to do them in this PR would make it much larger than it already is.  So to facilitate code review and testing I wanted to keep it simple.

But this has been sitting for a couple of months now so it would be nice to hear from @SuperSandro2000 in prospects for merging.

DivineDominion:
@dkerr64 From what I've heard it's _supposed_ to work with Dynv6; but maybe I'm reporting totally irrelevant stuff here and don't notice, sorry :)

SuperSandro2000:
> Requires review and testing by more than just me.

I don't really have the motivation to extensively test this right now which is why I didn't merge it yet.

tibbsbrookside:
Not sure if this helps but, just thought I'd say that I've tested this build with FreeDNS using both ifv6 and webv6 and it appears to work OK. I do get the following warnings but I'm not sure if they are related:

> WARNING:  '' is not a valid IPv4 or IPv6 address
WARNING:  found neither IPv4 nor IPv6 address
WARNING:  {HOST}: unable to determine IP address with strategy use=ip

adarobin:
I've used this successfully with CloudFlare.  IPv4 and IPv6 records are updated (assuming they already exist in CloudFlare).

```
daemon=600
syslog=yes
#ssl=yes

protocol=cloudflare
use=web
usev6=ifv6
ifv6=ens192
password=REDACTED_TOKEN
zone=redacted.net
minecraft.redacted.net
```

Neustradamus:
Any news?

dkerr64:
@DaveSophoServices thank you.  I'll be testing it.

PR #292, opened 2020-09-27
Fix Gandi LiveDNS update process.

deven:
I was trying to automatic dynamic DNS updates working with Gandi LiveDNS on my EdgeOS router, and encountered two separate issues:

1. There was a bug in the geturl() call inside the nic_gandi_update() routine, causing the update to fail because a hash reference was being passed into geturl(), which actually requires a list of key-value pairs instead.  This was a simple fix, just remove the braces from the geturl() call.

2. At least on my EdgeOS router, I was getting SSL certificate validation errors for "api.gandi.net", so I added a parameter to the geturl() call to disable the SSL hostname validation.  This isn't ideal, but it works around the problem successfully.

Since the magic EdgeOS commands to configure this are not obvious, here's how to configure Gandi LiveDNS on EdgeOS, if the WAN interface (to the Internet) is on eth0:
```
configure
delete service dns dynamic interface eth0
set service dns dynamic interface eth0 service custom-gandi protocol gandi
set service dns dynamic interface eth0 service custom-gandi login dummy
set service dns dynamic interface eth0 service custom-gandi password GANDI_API_KEY
set service dns dynamic interface eth0 service custom-gandi options zone=DOMAIN_NAME,ttl=5m
set service dns dynamic interface eth0 service custom-gandi host-name HOSTNAME
commit; save; exit
update dns dynamic interface eth0
show dns dynamic status
```
For the `host-name` parameter, a comma-separated list can be used to update multiple hostnames to the dynamic IP address, and `@` can be used as a hostname to update the IP address associated with the domain itself (which must be specified in the `zone=DOMAIN_NAME` parameter).  If the `ttl` option is not specified, it will default to 3 hours;
Gandi's minimum TTL supported is 5 minutes.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -5743,13 +5743,13 @@ sub nic_gandi_update {
         $url  = "https://$config{$h}{'server'}$config{$h}{'script'}";
         $url .= "/livedns/domains/$config{$h}{'zone'}/records/$hostname/$rrset_type";
 
-        my $reply = geturl({
+        my $reply = geturl(
             proxy   => opt('proxy'),
             url     => $url,
             headers => $headers,
             method  => 'PUT',
             data    => $data,
-        });
+        );
         unless ($reply) {
             failed("%s -- Could not connect to %s.", $h, $config{$h}{'server'});
             next;
--- a/ddclient.in
+++ b/ddclient.in
@@ -5744,11 +5744,12 @@ sub nic_gandi_update {
         $url .= "/livedns/domains/$config{$h}{'zone'}/records/$hostname/$rrset_type";
 
         my $reply = geturl(
-            proxy   => opt('proxy'),
-            url     => $url,
-            headers => $headers,
-            method  => 'PUT',
-            data    => $data,
+            proxy        => opt('proxy'),
+            url          => $url,
+            headers      => $headers,
+            method       => 'PUT',
+            data         => $data,
+            ssl_validate => 0,
         );
         unless ($reply) {
             failed("%s -- Could not connect to %s.", $h, $config{$h}{'server'});


SuperSandro2000:
We recently added ``ssl_validate`` which should solve your problem.

Also I suspect that your ca-certificates are just outdated cause api.gandi.net is valid on my machine. I don't want to hardcode a change that is specific to your setup. Can you remove that again?

deven:
> We recently added `ssl_validate` which should solve your problem.
> 
> Also I suspect that your ca-certificates are just outdated cause api.gandi.net is valid on my machine. I don't want to hardcode a change that is specific to your setup. Can you remove that again?
> 
> @dkerr64

I didn't like the idea of hardcoding it either, but I was trying to at least get it working.  I'll revisit it and try to make it work with the `ssl_validate` in the options and hopefully that will make it cleaner.  Give me a day or three to get back to it, don't merge just yet...

SuperSandro2000:
I am still not a fan of this. Can you please try updating your ca-certs first?

SuperSandro2000:
Closing due to inactivity. Feel free to comment and I can reopen the PR.

deven:
As I said at the start of this pull request, I found two **separate** issues.  They had nothing to do with each other.  I submitted both as a single pull request for simplicity, but that was obviously a mistake.  Since the two commits on this pull request were entirely independent of each other, I should have submitted two separate pull requests.  My bad!

Because I combined the two commits into a single pull requests and never found the time to follow up on the SSL certificate questions, the internal API fix in my first commit (0c6f4ef7c8383c502ea8efbbf09c5b2eda86e629) sat in limbo, even though the code was definitely broken and the fix was needed, as I clearly explained in the commit message based on my 33 years of Perl experience.  Eventually (almost 6 months later), someone submitted exactly the same fix for #314, successfully guessing based on other examples of `geturl()` calls in the code.  If I had made separate pull requests, this could've been merged months earlier, regardless of the SSL certificate question.  Oh well.

As it turns out, the certificate issue appears to be because IO::Socket::SSL on my EdgeOS router is so old that the Gandi certificate authority isn't supported.  I managed to update the ca-certificates for the operating system but it still didn't work in IO::Socket::SSL.  Rather than trying to upgrade IO::Socket::SSL, I tried a workaround: adding `curl=yes` to the options.  This fixed the problem, validating the Gandi certificate successfully via the `curl` command.  This makes my second commit moot; so this entire pull request can remain closed, thanks.

deven:
For reference, this was the commit message from the API fix:
```
Fix internal API error in geturl() call for Gandi.

In nic_gandi_update(), calling geturl() with a hash parameter is
incorrect usage, because geturl() calls either fetch_via_curl(@_) or
fetch_via_socket_io(@_), both of which contain "my %params = @_;" --
this causes a "Reference found where even-sized list expected" error
because @_ needs to be a list of key-value pairs to be assigned to
%params, not a reference to a hash containing those key-value pairs.

The fix is simply to remove the braces around the list of key-value
pairs from the geturl() parameters.
```

PR #52, opened 2018-01-19
dnsexit Support

sreknob:
Hi wimpunk -
I've hacked back in DNS exit support with this pull. Noticed it from https://sourceforge.net/p/ddclient/discussion/399428/thread/04e23ee6/ when I was gooling to try and make my dnsexit work. Took the patch that [Reuben Thomas](https://sourceforge.net/u/rrt/) updated from [danofsteel](https://sourceforge.net/u/danofsteel/) and did some further edits:
  - removed SSL as not supported
  - added in example to example config file
  - used the same regex that the official client uses to report update success
Hope this can make it into the master so I can easily deploy it in existing dockers with an update!
Thanks!

PR #310, opened 2021-02-05
Add support for dnsexit

rrthomas:
Patch from the one I posted to https://sourceforge.net/p/ddclient/discussion/399428/thread/04e23ee6


rrthomas:
Sorry, I overlooked #52.

PR #314, opened 2021-03-04
Fix geturl function call for Gandi

mchestr:
Was trying to setup Gandi dynamic DNS on my EdgeRouter which uses ddclient, however it wasn't working... upon further debugging it appears to be because these brackets are surrounding the function parameters, when they should be provided directly.

I am no perl expert but the other usages of `geturl` do not include the brackets, but after applying my fix I was able to successfully update my DNS record on Gandi.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,51 +53,51 @@ jobs:
       - name: distribution tarball is complete
         run: ./.github/workflows/scripts/dist-tarball-check
 
-  test-centos6:
-    runs-on: ubuntu-latest
-    container: centos:6
-    steps:
-      - uses: actions/checkout@v1
-      - name: install dependencies
-        run: |
-          yum install -y \
-              automake \
-              perl-IO-Socket-INET6 \
-              perl-core \
-              perl-libwww-perl \
-              ;
-      - name: autogen
-        run: ./autogen
-      - name: configure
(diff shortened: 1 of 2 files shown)


mchestr:
I see https://github.com/ddclient/ddclient/pull/292 also has this fix + more, but has been sitting for a few months, this PR should be good to go as is, and had no issues with certs

gerdneuman (approved):

gerdneuman:
Looks good, can this be merged?

SuperSandro2000:
> @SuperSandro2000 How do we re run the CI? I've disabled cento6 and 8, because they're not building right.

If the branch is new enough the actions are run.

deven:
Here's the commit message from my original patch with exactly the same fix, explaining why the braces needed to be removed:
```
Fix internal API error in geturl() call for Gandi.

In nic_gandi_update(), calling geturl() with a hash parameter is
incorrect usage, because geturl() calls either fetch_via_curl(@_) or
fetch_via_socket_io(@_), both of which contain "my %params = @_;" --
this causes a "Reference found where even-sized list expected" error
because @_ needs to be a list of key-value pairs to be assigned to
%params, not a reference to a hash containing those key-value pairs.

The fix is simply to remove the braces around the list of key-value
pairs from the geturl() parameters.
```

PR #324, opened 2021-04-05
modify Cloudflare documentation

krerkkiat:
It seems that ddclient check if the login field is equal to "token" to use the correct header for the API token.

https://github.com/ddclient/ddclient/blob/11a583b003920f8e15591813598b70061d1a4654/ddclient.in#L4764


panoti (approved):
It is very useful, I fixed my config by reading this instructions.

PR #353, opened 2021-08-17
Cloudflare Fix

a93h:
Cloudflare was returning values not being matched properly by the regex expression.
Numbers that were not Headers.
This fix or patch should resolve that issue, by only collecting one match to JSON relevant data from the response string.


arcnmx:
+1 this was required for the cloudflare protocol to not bail out for me (with no obvious error message I should add, which was confusing?)

PR #369, opened 2021-11-25
Trim non JSON part of payload from Cloudflare response to client

jrrk2:
I have noticed that the Cloudflare support does not work (for me) at the moment due to spurious HTTP data being returned before and after the wanted JSON data. This fix solved the problem for me, however I am not a Perl developer so somebody else should review it.


SuperSandro2000:
Already fixed by #353

PR #376, opened 2021-12-22
Updated example service file to use network-online.target

qs5779:
I finally took the time to determine why I received a connection error every morning when I first started my computer. In my opinion, this proposed change is a more correct way to control the ddclient service.

### DIFF
--- a/sample-etc_systemd.service
+++ b/sample-etc_systemd.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Dynamic DNS Update Client
-After=network.target
+After=network.target network-online.target
 
 [Service]
 Type=forking


sample-etc_systemd.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Dynamic DNS Update Client
-After=network.target
+After=network-online.target
```
SuperSandro2000, line 3:
```suggestion
After=network.target network-online.target
```

qs5779: Great suggestion, updated and pushed. I just now saw the button commit suggestion. Sorry.

Issue #334, opened 2021-05-11
WWW::Curl::Easy hangs


rrthomas:
When I was investigating #332, I needed to run with `-curl`. The first time I tried this, ddclient hung during the request. It was using `WWW::Curl::Easy`. Having uninstalled it, things worked fine with command-line curl.

Given that ddclient already has its own internal sockets implementation, is it really worth keeping WWW::Curl::Easy as well as the direct command-line interface?

dkerr64:
I implemented all the curl support, originally as command-line only.  But I was requested to add the WWW::Curl::Easy method as well.  As I look at the code again I see that it checks for and uses the module if it exists, there is no option to force command line method.  With hindsight I should have allowed for that.  But that is beside the point... do you have any debug/trace information that would tell us where the hang occurred ?

dkerr64:
> I don't think adding an option to force the command-line method is really an improvement: it still means that there are three code paths to maintain for the same functionality.

I agree... but FYI, the only way I got the command line curl accepted into the code was to also implement the module version.  I really needed the command version because the platform I use did not have the IPv6 socket modules.

dkerr64:
> It is at this point that it hangs, and Ctrl+C has no effect.
You're right, this doesn't give us many clues.  Is this the very first network request or has it successfully transmitted/received earlier requests?  When I get a chance I'll try and reproduce myself... I typically use Ubuntu, can you tell me what system/version the hang occurs on?

As for the Ctrl-C, ddclient catches those in order to manage a graceful exit which possibly explains why that is not breaking out of this hang.

rrthomas:
> You're right, this doesn't give us many clues.  Is this the very first network request or has it successfully transmitted/received earlier requests?

This is the first.

>  When I get a chance I'll try and reproduce myself... I typically use Ubuntu, can you tell me what system/version the hang occurs on?

Ubuntu 20.04 64-bit.

dkerr64:
> > I agree... but FYI, the only way I got the command line curl accepted into the code was to also implement the module version. I really needed the command version because the platform I use did not have the IPv6 socket modules.
> 
> This is maybe a decision to revisit?

All the history is in #233, you'll see it was quite a journey to get curl support added,  Personally I would remove the WWW::Curl::Easy code, but it is there because it was asked for.

rrthomas:
[...]

PR #401, opened 2022-03-21
Curl casy hangs #334

awalon:
commit 440a024bbd92b670daef8209897dcae6708ba1a0 could fix the issue with hanging WWW::Curl::Easy (#334)


awalon:
Hello @SuperSandro2000,
not sure if I did something wrong with my commits and/or branches as every change seems to be part of a single pull request, sorry for that.

SuperSandro2000:
I don't really want to add multi WAN support. I cannot easily test it and it makes the code I already barely understand not easier. My suggestion is to either use ``ifv[46]`` or writing a custom script and obtaining the IP via cmd.

PR #405, opened 2022-04-03
Update pid file path from /run/var/ to /run

gmelodie:
Was getting the following error while using the default systemd ddclient service:
```bash
PIDFile= references a path below legacy directory /var/run/, updating /var/run/ddclient.pid → /run/ddclient.pid;
```
Updating the path on the service file to `/run/ddclient.pid` fixes it.


SuperSandro2000:
IIRC that should be a warning and be none fatal.

PR #469, opened 2022-10-24
Cut 3.10.1 release

Denton-L:
In the 3.10.0 release, the `make check` tests fail because the version was not updated in configure.ac. Update both to v3.10.1 so that a new release can be made which passes the tests.


SuperSandro2000:
Fixed this in master by reducing the amount of versions number needing to be changed when updating.

PR #477, opened 2022-11-21
fix hetzner zone problem

lucaschoeneberg:
The query for zone_id cannot be done at /api/v1/records. Therefore, I put the query in dns_rec_id.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -6127,7 +6127,7 @@ sub nic_hetzner_update {
                 $config{$domain}{"status-ipv$ipv"} = 'failed';
 
                 # Get DNS 'A' or 'AAAA' record ID
-                $url = "https://$config{$key}{'server'}/records?$zone_id";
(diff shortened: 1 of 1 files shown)


schreibubi:
Hi @lucaschoeneberg ,

your change broke the hetzner update mechanism for me. v3.10.0 works fine, but latest master does not.

If you look at https://dns.hetzner.com/api-docs#tag/Records is is perfectly legal to query for the zone_id through records?zone.

So please either revert your change or update the following line in your commit accordingly (note the $zone_id eq $_->{zone_id}):
```
my ($dns_rec_id) = map { ($_->{name} eq $hostname && $_->{type} eq $type && $zone_id eq $_->{zone_id}) ? $_->{id} : ()} @{$response->{records}};
```

Joerg

lucaschoeneberg:
Sorry that it no longer runs for you. I'm going to take a look at it right now.

PR #480, opened 2022-12-08
Rewrite deprecated use of IO::Socket::INET6 to new solution with IO::Socket::IP

michal-josef-spacek:


SuperSandro2000 (approved):

michal-josef-spacek:
Could we merge?

PR #496, opened 2023-01-07
fix: CloudFlare configuration example

maximelafarie:
Current CloudFlare configuration makes it buggy. By removing useless backslashes and commas, it avoids weird errors


SuperSandro2000:
command and backslash are there to group settings together and allow to use multiple backends. They should generally work and the errors you have are most likely from a configuration mistake.

Issue #404, opened 2022-04-01
Missing IPv6 address in cache file ?


Highspeedy:
Hello,

in the cache file from my ddclient installation (dualstack IPv4 & IPv6) there is after an IP update no IPv6 cached.
The goal are unnecessary IPv6 updates on the DYNDNS provider every few minutes.

Is this a bug or ist it a configuration error of my ddclient.conf ?

#begin of file

#/etc/ddclient.conf

daemon=600
syslog=yes
#mail=root
#mail-failure=root
pid=/var/run/ddclient.pid
ssl=yes
cache=/var/cache/ddclient/ddclient.cache
protocol=dyndns2
usev4=webv4, webv4=ip4.ddnss.de/meineip.php/, webv4-skip='IP Address'
usev6=ifv6, ifv6=eth0
server=ddnss.de
login=xxxxx
password=xxxxxxxx
xxxxxxx.xxxxx.xx

#end of file

SuperSandro2000:
> Is this a bug or ist it a configuration error of my ddclient.conf ?

I would guess a bug but I don't have the capacity to investiagte this right now.

agowa338:
I'm using version 3.9.1 and the IPv4 isn't written to the cache either. It only wrote the hostname it checks. the `ip` attribute doesn't have any value in the cache.

tduboys:
Hi,
I had the same issue, you're using dyndns2 provider, and it was not working with ipv4/ipv6 logic
I tried to fix it on https://github.com/ddclient/ddclient/pull/502 and works correctly on my config, but I didnt tested every dyndns2 provider

PR #502, opened 2023-01-27
Update dyndns2 client to use new IPv4/IPv6 logic

tduboys:
Updating dyndns2 provider to work with the new IPv4/IPv6 logic, with wantipv4 and wantipv6 mechanism

tested with `usev4` and `usev6` config, separately and in the same conf.
if both provided, only 1 call is made to api, with ip separated by a comma, as described here https://help.dyn.com/remote-access-api/perform-update/


tduboys:
it will fix https://github.com/ddclient/ddclient/issues/404

PR #504, opened 2023-01-28
set use to disable if usev4 or usev6 is set

tduboys:
Hi,
Is it defined everywhere that `use` is deprecated and we need to use `usev4` and `usev6` options.
But, `use` have default value to `ip` https://github.com/ddclient/ddclient/blob/master/ddclient.in#L434

So, if we dont set `use` option to `no` and set `usev4` to something, the `sub update_nics` will try to define ip address with the IP module. it causes unwanted checks, maybe unwanted behavior, but mostly just unwanted warning on the log.

I suggest to disable `use` option if `usev4` or `usev6` is set.

Another option is to set `use` to `disabled` as default, but I think it could impact more people.


SuperSandro2000:
Sounds reasonable to me but I lack the time to properly test this on my own right now.

PR #512, opened 2023-02-06
Do not use @localstatedir@ for $cachedir.

comburent:
From <https://www.gnu.org/prep/standards/html_node/Directory-Variables.html>, it states that:
```
$(localstatedir) should normally be /usr/local/var, but write it as $(prefix)/var. (If you are using Autoconf, write it as ‘@localstatedir@’.) 
```

@localstatedir@ is usually /usr/local/var which unless explicitly overriden, will depend on @​prefix@.
This usage is problematic when the prefix is immutable (distributions such as Guix, NixOS and Fedora Silverblue) as the default value results in:
```
WARNING:  Failed to create cache file /gnu/store/k4nqw1g3qac7g6nj4xw036xw16n6zyp3-ddclient-3.10.0/var/cache/ddclient/.ddclient-real.cache: Read-only file system
```

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -49,7 +49,7 @@ sub subst_var {
 }
 
 my $etc = subst_var('@sysconfdir@', '/etc/ddclient');
-my $cachedir = subst_var('@localstatedir@', '/var') . '/cache/ddclient';
+my $cachedir = '/var/cache/ddclient';
 my $savedir = '/tmp';
 if ($program =~ /test/i) {
     $etc = '.';


comburent:
On second thought, the descriptions for @localstatedir@ and @​prefix@ [^1] describe them as if they're to be used as the standard '/etc' '/var' directories yet (confusingly) by default depend on @​prefix@.

My guess is that this is supposed to be controlled through the `./configure` script and that the current usage is correct. Might be worth mentioning for package maintainers.

[^1]: <https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html>

SuperSandro2000:
> Might be worth mentioning for package maintainers.

That's already the case here https://github.com/ddclient/ddclient#manual-installation . If that is missing something let me know.

> This usage is problematic when the prefix is immutable (distributions such as Guix, NixOS and Fedora Silverblue) as the default value results in:

It works on NixOS with the default autoconf steps https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/networking/ddclient/default.nix#L20

---

Also side note: I only really test ddclient with systemd and I plan to remove mentions for other init and process management systems because they are basically unmaintained.

PR #515, opened 2023-02-21
Merge configs for the same hosts instead of using the last one

tduboys:
Hi,

When having multiple configs for the same host, the current code always use the last one.
But on some cases, it's easyer to have multiple configs for the same host, one for ipv4, and another for ipv6.
Some systems like opnsense have one config per IP, so they use 2 configs for the same host if we want to use ipv4 and ipv6.

It should be fine if, when reading the config file, if a config already exists for a host, we merge the new and the already known into the same, instead of overwriting it.


tduboys:
I also updated Makefile to remove README.ssl as it's deleted from master
Feel free to remove this commit if you want to merge it separately

PR #520, opened 2023-02-26
Add support for Digital Ocean

tchebb:
Uses the Digital Ocean HTTP API. The added example explains the new protocol in detail.


SuperSandro2000:
I think you need to rebase to fix the ci checks.

tchebb:
Just rebased, but I still see a failure. That same failure exists on master, though (see [here](https://github.com/ddclient/ddclient/actions/runs/4416304797/jobs/7740408797)), so I don't think this change caused it.

SuperSandro2000:
Oh well, I don't care enough to fix that right now.

PR #542, opened 2023-06-01
dyndns2: fix multiline parsing and multiple host handling

fichtner:
As seen in the wild with DynDNS.com -- status '14' is being stored for the first host which is removed from @hosts ending up reading empty host for next line causing 'nochg' to be misplaced in an empty host.  The same likely applies for multi-host handling so expand to loop where writing to config and use $hosts when logging to catch all.

```
RECEIVE:  HTTP/1.1 200 OK
RECEIVE:  Date: Thu, 01 Jun 2023 06:59:38 GMT
RECEIVE:  Server: Apache/2.4.18 (Ubuntu)
RECEIVE:  Strict-Transport-Security: max-age=31536000 RECEIVE:  X-UpdateCode: n
RECEIVE:  Vary: Accept-Encoding
RECEIVE:  Content-Type: text/plain
RECEIVE:  Accept-Ranges: none
RECEIVE:  X-User-Status: vip
RECEIVE:  Connection: close
RECEIVE:  Transfer-Encoding: chunked
RECEIVE:
RECEIVE:  14
RECEIVE:  nochg 192.168.178.20
RECEIVE:  0
RECEIVE:
```

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -4194,30 +4194,38 @@ sub nic_dyndns2_update {
                 # bug #10: some dyndns providers does not return the IP so
                 # we can't use the returned IP
                 my ($status, $returnedips) = split / /, lc $line;
-                my $h = shift @hosts;
 
-                $config{$h}{'status'} = $status;
-                $config{$h}{'status-ipv4'} = $status if $ipv4;
-                $config{$h}{'status-ipv6'} = $status if $ipv6;
+                foreach my $h (@hosts) {
+                    $config{$h}{'status'} = $status;
+                    $config{$h}{'status-ipv4'} = $status if $ipv4;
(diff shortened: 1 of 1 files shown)


fichtner:
@SuperSandro2000 thanks. the same likely applies to other update functions and/or addresses a few open issues. I might take a look at duckdns when 3.11.0 is out to see if it requires similar treatment. Got any ETA on 3.11.0? OPNsense is currently using a modified version of 3.10.0 with a few extra patches and I know that FreeBSD ports is holding off the 3.10.0 version for these reasons.


Cheers,
Franco

SuperSandro2000:
> Got any ETA on 3.11.0?

Nothing is planned right now.

Issue #552, opened 2023-07-19
3.11 release plan & Maintenance discussions


rrthomas:
[Copied from rrthomas/ddclient#8]

A quick backwards-compatible release with essential bug-fixes and functionality enhancements only. To-do list:

- [x] Fix caching problems (rrthomas/ddclient#3)
- [x] Update changelog
- [x] Contact downstream package maintainers? (Change of maintainers, change of curl dependency to be required, ...)
- [x] Merge back into the original repo
- [x] Set up signed tags (and commits)
- [ ] Update docs repo with latest changes

LenardHess:
- [ ] Update docs repo with latest changes

LenardHess:
Also - the link to the caching problems issue wasn't copied over correctly

rrthomas:
Thanks @LenardHess, I've added your item and fixed the link.

Majiir:
What's left to do here? It looks like most of the TODOs relate to switching repos which is no longer necessary.

rrthomas:
@Majiir I have ticked the box for the item related to changing repos. The other two still need doing.

LenardHess:
@rrthomas Since we got quite a few provider updates and new implementations pending now - I'd say we do these steps
1. Get v3.11.0 out the door
2. Work on v3.11.1:
 a. Pull in provider changes
 b. Check in with downstream maintainers (including letting some time pass in case anything pops up delayed)
3. Get started with deprecation/cleanup discussions for v4.0.0 (iirc. I had thoughts about your cleanup branch but haven't bothered writing them down yet)

TinfoilSubmarine:
FWIW, I've got the master branch packaged and working on Void so once a release is made Void should be all set.

rrthomas:
> @rrthomas Since we got quite a few provider updates and new implementations pending now - I'd say we do these steps
> 
>     1. Get v3.11.0 out the door
> 
>     2. Work on v3.11.1:
>        a. Pull in provider changes
>        b. Check in with downstream maintainers (including letting some time pass in case anything pops up delayed)
> 
>     3. Get started with deprecation/cleanup discussions for v4.0.0 (iirc. I had thoughts about your cleanup branch but haven't bothered writing them down yet)

This all sounds good to me. Thanks for working this all out!
 
> For 1. in particular I'd still opt to pull any currently open pull request that fixes provider _regressions_. I.e. the porkbun fix won't qualify as porkbun is newly added with v3.11.

[...]

PR #557, opened 2023-07-31
porkbun: rework logic to allow for simultaneous update of IPv4 and IPv6

TinfoilSubmarine:
Just had a look at the diff and it was not too obvious what I did, so here's the gist:

- Update porkbun logic to properly handle dual updates to IPv4 and IPv6 at one time
- Also slightly deduplicated the porkbun logic by unifying the ipv4 and ipv6 logic using `$ipv`

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -7076,23 +7076,25 @@ EoEXAMPLE
 sub nic_porkbun_update {
     debug("\nnic_porkbun_update -------------------");
 
-    ## update each configured host
-    ## should improve to update in one pass
     foreach my $host (@_) {
-        my ($sub_domain, $domain);
-        if ($config{$host}{'on-root-domain'}) {
-            $sub_domain = '';
-            $domain = $host;
-        } else {
-            ($sub_domain, $domain) = split(/\./, $host, 2);
-        }
-        my $ipv4 = delete $config{$host}{'wantipv4'};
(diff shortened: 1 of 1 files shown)


LenardHess:
From a quick glimpse of the old code - it doesn't seem like simultaneous IPv4 and IPv6 was broken?

Cleanup/Refactor looks good to me though :+1: 

I assume you're using porkbun and tested this?

TinfoilSubmarine:
> From a quick glimpse of the old code - it doesn't seem like simultaneous IPv4 and IPv6 was broken?

That's what I was confused about, but I think that the `next` calls were messing it up. I didn't verify that this was true though.
> Cleanup/Refactor looks good to me though 👍
> 
> I assume you're using porkbun and tested this?
Yes, testing it out directly on the Porkbun API with multiple domains, root domain, etc.

TinfoilSubmarine:
OK, sorry for another set of revisions to the commit, but I didn't like the addition of the extra function when the same thing could be accomplished with a foreach. Logic remains the same.

LenardHess:
Thanks for the update!

We will leave this pull request open until v3.11.0 has been released (see https://github.com/ddclient/ddclient/issues/552#issuecomment-1675925270 for details). It will get pulled into v3.11.1

LenardHess:
With v3.11.0 released, please rebase ontop of current master.

Issue #587, opened 2023-11-02
Incorrectly reports using non-existent IP version 0


TinfoilSubmarine:
There is a confusing line of code at https://github.com/ddclient/ddclient/blob/afa127525380d8c3e19b2046bca6843346b1ab0d/ddclient.in#L2573

Which outputs at https://github.com/ddclient/ddclient/blob/afa127525380d8c3e19b2046bca6843346b1ab0d/ddclient.in#L2642 as well as https://github.com/ddclient/ddclient/blob/afa127525380d8c3e19b2046bca6843346b1ab0d/ddclient.in#L2604

The problem with this is that when there is an error, the script outputs something like the following:
```
WARNING:  CURL error (6) Couldn't resolve host. The given remote host was not resolved.
WARNING:  curl cannot connect to https://api.gandi.net/ using IPv0
FAILED:   @ -- Could not connect to api.gandi.net.
```
This could lead users to falsely believe that the issue is that we are trying to use a non-existent IP version.

I'm confused about what the intent of setting it to 0 is. Is it to indicate that we are supposed to use all available protocols (meaning 4 and 6)? **If that is the case**, wouldn't it be better to do something like
```
my $ipversion = ($params{ipversion}) ? int($params{ipversion}) : "4/6";
```
instead then? Obviously would need to string-ify the only 4 and only 6 cases, but this would better indicate to the user what is going on.

And if this isn't the case, if we don't read an IP version number from `$params{ipversion}`, shouldn't we handle that error instead of continuing with 0?

LenardHess:
The `ipversion` parameter to `geturl()` is only used to force IPv4/6 respectively. So neither being forced (ipversion = 0) is a valid usecase of the function. The log messages should probably be reworded to omit the "using IPv*" part if neither IP version option is forced.

PR #593, opened 2023-11-08
don't print erroneous IP version

TinfoilSubmarine:
Fixes https://github.com/ddclient/ddclient/issues/587


LenardHess:
Merged (after a local rebase) - that seems to have broken Github's awareness of the merge. I take it i shouldn't do it this way next time :thinking:

Issue #597, opened 2023-11-14
`use=web` default no longer works as https://checkip.dyndns.org is no longer reachable anymore 


wheelerlaw:
https://checkip.dyndns.org is not reachable (the plain HTTP one is though). The default option for `use=web` should be changed to one that is working.

LenardHess:
ddclient will choose http/https depending on the URL by default.
There is the option to force every connection to use HTTPS - the `-ssl` option.

Did your configuration have `https` in the URL or did you set the `-ssl` parameter?

bryanhorstmann:
I can confirm I'm seeing the same thing on `3.11.x` with `checkip.dyndns.com`. Here are my tests:

With the following config (from `3.10`):
```
ssl=yes
use=web, web=checkip.dyndns.com/, web-skip='IP Address'
```
I get the following:
```
WARNING:  CURL error (28) Operation timeout. The specified time-out period was reached according to the conditions.
WARNING:  curl cannot connect to https://checkip.dyndns.com using IPv4
WARNING:  Could not determine an IP for <my domain>
```

With this config:
```
ssl=yes
usev4=webv4, webv4=checkip.dyndns.com/, web-skip='IP Address'
```
I get the following:
```
WARNING:  CURL error (28) Operation timeout. The specified time-out period was reached according to the conditions.
WARNING:  curl cannot connect to https://checkip.dyndns.com using IPv4
WARNING:  Could not determine an IP for <my domain>
```

With this config:
```
ssl=yes
usev4=webv4, webv4=https://checkip.dyndns.com/, web-skip='IP Address'
```
I get the following:
```
WARNING:  CURL error (28) Operation timeout. The specified time-out period was reached according to the conditions.
WARNING:  curl cannot connect to https://checkip.dyndns.com using IPv4
WARNING:  Could not determine an IP for <my domain>
```

With this config:
```
ssl=no
usev4=webv4, webv4=http://checkip.dyndns.com/, web-skip='IP Address'
```
I get the following:
```
WARNING:  updating <my domain>: nochg: No update required; unnecessary attempts to change to the current address are considered abusive
```

If I revert back to `3.10` with my original config file it works as expected.

BraxtonLowers:
I recently observed the same issue. Default ddclient.conf sets https as the only option because of the config line
`ssl=yes                                  # use ssl-support.  Works with`
[...]

PR #608, opened 2023-12-16
If present, give URL protocol precedence over other SSL settings

mmogilvi:
This is basically a minor adjustment to https://github.com/ddclient/ddclient/pull/482 , updated for the new 3.11 curl-based geturl().  The original PR objections seemed to be focused on lack of time to properly understand the nuances of the old behavior rather than any realistic objection to the only-slightly-different new behavior.

My setup broke for the 3.10 update, with no easy config-only fix (see below).  Now Gentoo just updated to 3.11.1, and the 3.10 version of my fix/patch can't be applied, hence this newer patch.  This patch doesn't directly restore 3.9, but it is now possible to make a trivial configuration change for situations like this, by just including the protocol as part of each individual broken URL.

Key points:
  - I think it is basic good citizenship to prefer to extract your IP address from your local router, rather than asking a third party service unnecessarily.  Especially when querying it on a regular basis.
  - Many (most?) routers probably don't try to support an SSL web interface (https), probably because handling server certificates for a private service like this is not trivial (expiration, private URL/IP address changes, etc).  At least mine doesn't.
  - Best to use SSL wherever possible, though.  Especially for the actual update to the external ddns service you are using.
  - However, as near as I could tell (at least for 3.10), there was no way to enable SSL for the update without also forcing SSL for the "get IP address".  (It was somehow working in 3.9, although I'm not sure if it was using SSL at all.  3.10 forced SSL more widely (by default), and broke by setup.)
  - Related: At least a few other bugs mention difficulties using plain http for getting IP address...
  - The existing separate SSL setting is still used when the "URL" is incomplete and lacks a protocol, so the behavior change is minimal.

FUTURE ideas:
 - Perhaps this should be documented somehow.  (Although the documentation isn't great in general.)
 - Are there more code paths where similar "give precedence to URL protocol" should be changed?
 - Maybe some of the commented URL's in the default configuration file should include the protocol (http or https), especially if it is known that only one actually works.
 - Relatively few default config (commented) URLs currently specifies a protocol directly, but maybe "lack of explicit protocol in URL" should be deprecated because it leaves the nuances of configuring things more confusing...  (Protocol appears to be included in several embedded URLs now, though, based on a quick grep of the main script.)

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -2578,21 +2578,27 @@ sub geturl {
     my $reply;
     my $server;
     my $use_ssl = 0;
-    my $force_ssl = 0;
     my $protocol;
     my $timeout  = opt('timeout');
     my @curlopt = ();
     my @header_lines = ();
 
-    ## canonify proxy and url
-    $force_ssl = 1 if ($url =~ /^https:/);
+    ## canonify use_ssl, proxy and url
+    if ($url =~ /^https:/) {
+        $use_ssl = 1;
+    } elsif ($url =~ /^http:/) {
+        $use_ssl = 0;
+    } elsif ($globals{'ssl'} && !($params{ignore_ssl_option} // 0)) {
+        $use_ssl = 1;
+    } else {
+        $use_ssl = 0;
+    }
+
     $proxy  =~ s%^https?://%%i if defined($proxy);
     $url    =~ s%^https?://%%i;
     $server = $url;
     $server =~ s%[?/].*%%;
     $url    =~ s%^[^?/]*/?%%;
-
-    $use_ssl = 1 if ($force_ssl || ($globals{'ssl'} && !($params{ignore_ssl_option} // 0)));
 
     $protocol = ($use_ssl ? "https" : "http");
 


LenardHess:
Looking at the old behavior and this patch, this is what is happening:
- Existing behavior:
  - If URL has https://, use HTTPS
  - If `ssl` is set, use HTTPS, except when disabled via `ignore_ssl_option`
  - Otherwise use HTTP
- New behavior:
  - If URL has https://, use HTTPS
  - **If URL has http://, use HTTP**
  - If `ssl` is set, use HTTPS, except when disabled via `ignore_ssl_option`
  - Otherwise use HTTP

The only changed behavior is therefore for `http://` URLs with `ignore_ssl_option=0` and `ssl=1`. Those would use HTTP instead of HTTPS with this change.

I like this addition, as it makes the path to forcing HTTP more clear in my opinion (use `http://` in the URL). This would also allow dropping the `ignore_ssl_option` parameter in `geturl` calls by converting the URL passed to a `http://` URL.

As mentioned in #597 - I'm inclined to add a "Disable HTTP"/"Force SSL everywhere" option to allow users to opt out of HTTP all together by either force-upgrading to HTTPS or by logging a warning/error on HTTP. The implication of such an option would be that the user then find any HTTP usage and fix it (Move away from HTTP-only services/providers).
But this will most likely have to wait until more urgent stuff is done.

Drag0nFly:
Apparently I'm missing something, as this still does not work for me.

Still getting errors that it cannot determine an IP address, even after applying https://github.com/ddclient/ddclient/pull/608
and having 

`use=web, web=http://checkip.dyndns.org/` set above the hosts to be updated (and no other use statement present)

Running manually and via the daemon appears to cause different results, as a manual run prints `skipping update from <nothing> to <correct IP address>` whereas the daemon only prints that it `could not determine an IP` in the syslog.

This is really breaking the functionality of ddclient (which I've used for more than 15 years). Hopefully there's an easy fix (such as switching to a default checkip service which actually works with SSL)

I spent the better part of two days trying to resolve this, updating to various versions, applying PRs, and modifying config, and still came up short.

PR #614, opened 2024-01-04
Update Mythic Beasts module (mythicdyn) in ddclient.in

retep:
The mythicdyn module is modified so that it will update _either or both_ V4 _and/or_ V6 addresses depending upon which specific address parameters have been defined in the config file.  The module monitors:
  `cmdv[46] fwv[46] usev[46] ipv[46] & ifv[46]`
If required, both addresses will be updated in a single invocation.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -5835,11 +5835,13 @@ www.mythic-beasts.com.
 
 Configuration variables applicable to the 'mythicdyn' protocol are:
   protocol=mythicdyn           ##
-  ipv6=no|yes                  ## whether to set an A record (default, ipv6=no)
-                               ## or AAAA record (ipv6=yes).
   login=service-login          ## the user name provided by the admin interface
   password=service-password    ## the password provided by the admin interface
   fully.qualified.host         ## the host registered with the service
+
+Note: this module examines the following v4 & v6 parameters:
+    cmdv[46] fwv[46] usev[46] ipv[46] & ifv[46]
+  and will set either or both V4 and/or V6 addresses as required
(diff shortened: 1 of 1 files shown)


LenardHess:
Instead of reaching into the config and checking the cmd/if/ip/... stuff - please just check for the presence of `wantipv4` and `wantipv6` instead to determine which IP versions to update

retep:
> Instead of reaching into the config and checking the cmd/if/ip/... stuff - please just check for the presence of `wantipv4` and `wantipv6` instead to determine which IP versions to update

Thank you for  your suggestion.  I had found `wantipv[46]` when reading the source code about 6 weeks ago and I thought I'd checked their relevance, then.  However, i've checked again and I'm happy to include your suggestion in the rework, and it's made it a lot simpler (shorter).  I've also taken the opportunity to set `status-ipv[46]` as appropriate for the cache.  I've left `status` unchecked as I see there there's a "TODO" flag in the source suggesting it could/should be removed.  Please let me know if you think this needs changing.

retep:
Apologies for closing this PR and creating a new one.  After much heading banging with Github, it was the easiest way I could find to get a PR back in line with the master.

LenardHess:
Ok - for future reference: To update a pull request, simply update the branch it was made on (i.e. retep:retep-patch-1 in this case). You can push new commits to it to add them to the PR or you can force-push the branch to fully update the commits of the PR.

retep:
Thanks for your patience and the guidance whiich is much appreciated.

Issue #613, opened 2024-01-03
Rework the Mythic Beasts implementation


retep:
Rework the Mythic Beasts implementation so that:

1. IPV6 requirements are correctly detected
2. Permit one invocation to update both IPV4 and IPV6 addresses (pseudo simultaneous)

This has been tested with the built-in service within OPNsense (v23.7.10_1) with a minor tweak.

LenardHess:
Hi,

this reads like you meant to do a pull request instead of an issue?

retep:
Indeed, you are correct.  With assistance (thank you @rrthomas ) I have created a pull request (my first on Github - please be gentle).

rrthomas:
Thanks for the PR @retep, closing this issue in favour of #614.

PR #616, opened 2024-01-11
Update Mythic Beasts mythicdyn module in ddclient.in

retep:
The mythicdyn module is modified so that it will update either or both V4 and/or V6 addresses depending upon which specific address parameters have been defined in the config file. The module examines the wantipv4 and wantipv6 parameters. If required, both addresses will be updated in a single invocation.


LenardHess:
(Previous PR here: #614, previous issue here: #613 )

retep:
Thank you.

Issue #602, opened 2023-11-30
Gandi has deprecated API keys


mpartel:
Gandi now recommends [Personal Access Tokens](https://api.gandi.net/docs/authentication/) instead of API keys.

Currently, the authorization header looks like `Authorization: Apikey <password>`
With Personal Access Tokens, it should be `Authorization: Bearer <password>`

I tried making this change in ddclient, and it works, but of course just changing it would break things API key users.

I did not find a timeline for when Gandi will remove API keys.

G-eos:
It seems that this is effective now :-(

LenardHess:
@G-eos here is an untested pull request. Can you check whether this works?

G-eos:
Yes it works well :) thanks !

catwith1hat:
API keys still work, therefore #625 broke all your existing Gandi users. I suggest to revert that PR and adopt #626 to allow both, Bearer and Apikeys.

G-eos:
Crap, I am sorry for the bad info. And this breaking existing user ! This did not work for me. May this was another issue. I raised question to Gandi support today to get an official info on when it is planned to stop it.
Note that the API Key is already announced as deprecated.

catwith1hat:
@G-eos API keys are deprecated for many years now, and yet they still work. At least they did 3 weeks ago, when I migrated away from them for other reasons. Note that API keys require Gandi's "Classic DNS" server. If you are on LiveDNS, you need a PAT for some reason. Luckily, supporting both is trivial.

(I don't really understand Gandi's policy here. What difference does it make if I add an opaque string prefixed with "Bearer" to adding an opaque string prefixed with "Apikey"? That's one and the same to me.)

LenardHess:
@G-eos @mpartel are you available to test #636 ? Ideally with both an API key and a personal access token :wink:

mpartel:
:ballot_box_with_check: Tested with personal access tokens.

I don't have time to properly test with API keys, but the PR looks like it should work, and turning off `use-personal-access-token` while using a personal access token did result in a failure, as it should.

rhansen:
Some more possible approaches:
  * `password: my-gandi-api-key` for an API key and `personal-access-token: my-gandi-personal-access-token` for a personal access token.  If both are given, ignore `password` (with a warning).
[...]

PR #626, opened 2024-02-04
Gandi: Add support for personal access tokens.

catwith1hat:


LenardHess:
Hello,
the gandi API update has already been done in this commit (a57cb3b). See also #602 for more

catwith1hat:
The comments in #602 wrong. API keys still work. So you just surprise broke all your Gandi users with the other backward incompatible PR. I would revert the other PR and merge this one before you cut a new release. But that's your choice.

LenardHess:
In that case I'll get #626 reverted then and this one looked at, thanks for the info :+1:

LenardHess:
I do not want the password string to have a magic prefix - especially considering that when (if?) gandi kills the support for API keys, we want to remove support for the old method and make the PAT method the only one available. Having those updates affect the details on how to format the password entry is messy.

I tried a different route in #626 - using a separate config bool variable.

rhansen:
See <https://github.com/ddclient/ddclient/issues/602#issuecomment-2002133559> for additional alternative approaches.

LenardHess:
Merged #636, closing this one

PR #627, opened 2024-02-05
porkbun code not handling root domain

etkal:
When specifying the root domain the URL portion being sent to porkbun was:
  api/json/v3/dns/retrieveByNameType/example.com/A/example.com
However, per the API specification if it is not a subdomain it needs to be of the form:
   api/json/v3/dns/retrieveByNameType/example.com/A/

I added code to test for a match on the domain root, and also fixed the check for the specified host not being a subdomain of the specified domain root.

In addition I found that calls to verbose() require a prefix string, which was not specified in all calls.


LenardHess:
Thanks for the contribution :+1:

Issue #501, opened 2023-01-27
Error 


mbacso:
Hi,
I have an error updating to keysystems
I am using ddclient on OpnSense v 23.1 and ddclient installed from plugin default repository v1.11.

The first error I got seemed to be a missing parameter in the config file, outputing the following message:
WARNING:  skipping host: <my-domain>: 'server=1' is an invalid fully qualified host name and optional port number.

The original ddclient.conf, after making all settings on the GUI looks like this:
syslog=yes                  # log update msgs to syslog
pid=/var/run/ddclient.pid   # record PID in file.
verbose=yes
ssl=yes

use=if, if=pppoe0, \
protocol=keysystems, \
login=<my-domain>, \
password=<my-password> \
<my-domain>

I did some research and simpy added a line to the config ending up like this
...
use=if, if=pppoe0, \
protocol=keysystems, \
server=dynamicdns.key-systems.net, \
login=<my-domain>, \
password=<my-password> \
<my-domain>

That seemed to help somewhat and I got another error message.
This is the output I get when running ddclient -daemon=0 -debug -verbose -noquiet

INFO:     KEYSYSTEMS setting IP address to <my-public-ip> for -<my-domain>
Use of uninitialized value $_[0] in list assignment at /usr/local/sbin/ddclient line 2413.
Use of uninitialized value $url in pattern match (m//) at /usr/local/sbin/ddclient line 2427.
Use of uninitialized value $url in substitution (s///) at /usr/local/sbin/ddclient line 2429.
Use of uninitialized value $server in substitution (s///) at /usr/local/sbin/ddclient line 2431.
Use of uninitialized value $url in substitution (s///) at /usr/local/sbin/ddclient line 2432.
Use of uninitialized value $url in substitution (s///) at /usr/local/sbin/ddclient line 2432.
DEBUG:    proxy    = <undefined>
DEBUG:    protocol = https
Use of uninitialized value $_[0] in sprintf at /usr/local/sbin/ddclient line 2163.
DEBUG:    server   =
DEBUG:    url      =
DEBUG:    ip ver   =
Use of uninitialized value $peer in substitution (s///) at /usr/local/sbin/ddclient line 2450.
Use of uninitialized value $peer in pattern match (m//) at /usr/local/sbin/ddclient line 2451.
Use of uninitialized value $peer in pattern match (m//) at /usr/local/sbin/ddclient line 2451.
[...]

PR #629, opened 2024-02-14
Fix for getting url for keysystems

code-chicken:
Updating keysystems won't work without this fix as described here: https://github.com/ddclient/ddclient/issues/501#issuecomment-1895708060


LenardHess:
Thanks for the contribution, merged :+1:

PR #639, opened 2024-03-07
User-friendly pre-release version strings

rhansen:
(multiple commits; please do not squash)

Perl version strings are flawed in a few ways.  Convert them to user-friendly strings when printed so that Git tags and tarball names are easier for downstream distributions to work with.

Also tell Autoconf to read the version from `ddclient.in` to avoid the need to maintain the same version string in two different places.


LenardHess:
Quick FYI: I haven't had the time to look at this (or the general build steps for that matter) so far.

rhansen:
I'm going to go ahead and merge this, though I would appreciate it if someone other than me would look over it to confirm that the approach I took is reasonable.  I can easily revert the merge commit and re-do this PR if anyone has any suggestions for improvement.

PR #641, opened 2024-03-13
karn2544Create google.yml

thicha10:


rhansen:
I don't understand how this is relevant to the ddclient project or its users.  Can you provide more context/motivation?

PR #646, opened 2024-03-28
Fix DonDominio service response treatment

toniwiki:
This PR fixes a bug in handling the response from DonDominio service.

According to [DonDominio documentation](https://dondominio.dev/es/dondns/docs/api/#respuesta-plaintext) when the DNS update has succeeded the service returns "OK" and, if the IP has changed, it returns an additional line with "IP:x.x.x.x.x". Also, in practice, the service always returns this line with IP as the last line, which causes ddclient to report the update as a failure because the last line is not the string "OK".

This PR corrects this behaviour.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -6935,7 +6935,7 @@ sub nic_dondominio_update {
 
         my @reply = split /\n/, $reply;
         my $returned = pop(@reply);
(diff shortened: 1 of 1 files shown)


LenardHess:
Merged, thanks for the contribution :+1:

PR #653, opened 2024-04-21
Add 'emailonly' client to send status emails without needing a DDNS host

TV4Fun:
This adds a protocol to send status updates without needing a Dynamic
DNS host. This is useful if you don't have a DDNS host but want to be
updated when the IP of a machine changes. Because of how ddclient is
written, it requires setting a host name, but doesn't actually do
anything with it. Let me know if there's a better way to handle this.


TV4Fun:
Sorry, did not realize my branch was 2 years old. I'll submit a separate PR with this fixed.

PR #654, opened 2024-04-21
Add 'emailonly' client to send status emails without needing a DDNS host

TV4Fun:
This adds a protocol to send status updates without needing a Dynamic
DNS host. This is useful if you don't have a DDNS host but want to be
updated when the IP of a machine changes. Because of how ddclient is
written, it requires setting a host name, but doesn't actually do
anything with it. Let me know if there's a better way to handle this.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -41,6 +41,8 @@ repository history](https://github.com/ddclient/ddclient/commits/master).
   * The second and subsequent lines in a multi-line log message are now prefixed
     with a `|` character.
     [#676](https://github.com/ddclient/ddclient/pull/676)
+  * `emailonly`: New `protocol` option that simply emails you when your IP
+    address changes.  [#654](https://github.com/ddclient/ddclient/pull/654)
 
 ### Bug fixes
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,6 +65,7 @@ handwritten_tests = \
 	t/builtinfw_query.pl \
 	t/get_ip_from_if.pl \
 	t/geturl_connectivity.pl \
+	t/interval_expired.pl \
 	t/is-and-extract-ipv4.pl \
 	t/is-and-extract-ipv6.pl \
 	t/is-and-extract-ipv6-global.pl \
--- a/ddclient.conf.in
+++ b/ddclient.conf.in
@@ -391,3 +391,9 @@ ssl=yes					# use ssl-support.  Works with
 # password=ddns_password
 # redirect=2
 # example.com
+
+##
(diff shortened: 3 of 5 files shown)


rhansen:
Thank you for the contribution!  I rebased and added a handful of changes.  If you approve, I'll squash and merge.

TV4Fun (commented):
LGTM. Tested on my machine and it works properly. Thank you :). One very minor quibble, I like the idea of adding some indentation to host lines after the header, but with the `SUCCESS:` prefix before it, the result looks a little odd. See my screenshot.
Screenshot 2024-05-20 at 6 58 03 AM

rhansen:
By "a little odd" do you mean that there is too much indentation?  Or that it would be better to omit the `SUCCESS:` prefix?  Feel free to tweak it yourself.

I set it to right align with a wide field so that domains would align like this: 

```
             short.example.com  192.0.2.5
  reallyreallylong.example.com  192.0.2.19
```

TV4Fun:
Yes, I think it looks better without the `SUCCESS:` prefixes. Pushed a change to do that.
Screenshot 2024-05-20 at 9 28 39 PM

rhansen:
I rebased the branch and updated it to use the new `logmsg` function I added in #676.  I also tweaked the commit message wording.  Please take a look and let me know what you think.  Please push any changes you feel are appropriate.

TV4Fun:
@rhansen, I am not sure why, but there seems to be a bug introduced in this version—and the previous one—that was not in my original submission, that if the IP address has not changed since the last update, the update causes ddclient to report

> WARNING:   found neither IPv4 nor IPv6 address
>WARNING:   Could not determine an IP for <hostname>

I am investigating.

rhansen:
There was a bug in my infinite duration commit.  I force-pushed a fix; please try again.

However, the error looks unrelated to this PR.  Maybe something I recently changed on `master` introduced a bug?  I can't reproduce your error, though the test config I'm using is very simple:

```shell
cat <<\EOF >./test.conf
protocol=emailonly \
usev4=ipv4 ipv4=192.0.2.1 \
usev6=ipv6 ipv6=2001:db8::1 \
host.example.com
EOF
make && ./ddclient --file=./test.conf --cache=./test.cache
```

TV4Fun:
Yes, I have been running this branch now for 24 hours and have not had the issue again, so it looks like it has been fixed. Everything else is LGTM.

rhansen:
Thanks for testing and thanks for the contribution!

TV4Fun:
Not at all, thank you for your improvements!

PR #722, opened 2024-07-29
inwx: Add missing `hostname` argument to update URL

rhansen:
@Starkstromkonsument Does this look correct to you?  If so, would you please test it?

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -6038,15 +6038,14 @@ sub nic_inwx_update {
         delete $config{$_}{'wantipv6'} for @hosts;
         info("$hosts: setting IPv4 address to $ipv4") if $ipv4;
         info("$hosts: setting IPv6 address to $ipv6") if $ipv6;
-        my $url = "$groupcfg{'server'}$groupcfg{'script'}?";
-        $url .= "myip=$ipv4" if $ipv4;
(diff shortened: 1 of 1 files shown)


Starkstromkonsument:
Hi, INWX does not need / support the `hostname` argument. They only support these arguments in their Update-URL: `https://dyndns.inwx.com/nic/update?myip=<ipaddr>&myipv6=<ip6addr>`.

Background: Each hostname has its own DynDNS-Account with an own combination of username/password. Here is a screenshot of their add account dialog:

IMG_20240729_072604

Starkstromkonsument:
The hostname can not be changed:

IMG_20240729_073335

rhansen:
Thanks for the info!  I'll add a comment.

I really wish they would document their API...

Starkstromkonsument:
> I really wish they would document their API...

Me too! But they actually have an alternative [XML-RPC / JSON-RPC API](https://www.inwx.de/en/offer/api) for complete DNS management, that is properly documented. It could be used for DynDNS too, but that would be way more complicated and exceeds my capabilities.

There are some ready to use implementations for it e.g. [libdns](https://github.com/libdns/inwx). Maybe adding support for an interface to libdns or something similar would be something for a future roadmap for ddclient?

PR #724, opened 2024-07-29
Correct NoIP example

vladmovchan:
An extra comma in the example causes `401 Unauthorized` / `badauth`


rhansen:
Thanks!  We should fix config file parsing so that a trailing comma is treated as a separator, not part of the password.

vladmovchan:
Thank you!

PR #729, opened 2024-08-03
Removed 5m min-interval from changeip protocol

tamasbakos:
Removed the min-interval that was set to 5 minutes in changeip, because in some usecases it is needed to be able to change the IP and change it back quickly. According to my tests, changeip has no problem updating every 30 seconds, which is the default min-interval value in ddclient, and would be applied after this change.


rhansen:
Thanks!  Most DDNS services have a minimum duration requirement but I couldn't find anything in their documentation so  LGTM.

Issue #745, opened 2024-09-12
`escaping \# comment markers is not portable` warning


chenrui333:
👋 while regression building 3.11.2 release, seeing some build failure as below:

```
automake: warnings are treated as errors
Makefile.am:20: warning: escaping \# comment markers is not portable
autoreconf: error: automake failed with exit status: 1
ERROR: 'autoreconf -fviW all' failed
```

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/10821142464/job/30022616695

ripdog:
This is breaking the build on the AUR package as well.

dimpase:
and the Gentoo package

rhansen:
Fixed by commit a12398c315b9b909e57e87acf9fd3a15a0b3e213

barracuda156:
Also affects macOS.

PR #746, opened 2024-09-12
build: disable automake treating warnings as error

chenrui333:
dont see any usefulness with automake treating warnings as errors. this is the same patch suggested in https://github.com/chimera-linux/cports/commit/ee4dc57b4316a390760b9bb3f0211917c0baf324

fixes https://github.com/ddclient/ddclient/issues/745


rhansen:
I like the `-Werror` so I fixed the build failure a different way:  a12398c315b9b909e57e87acf9fd3a15a0b3e213

PR #747, opened 2024-09-21
Add snap packaging support

kubiko:
Adding support for snap packaging.

As snap packaging is supported on multiple linux distributions, this would allow running the latest `ddclient` even on the older releases. While still getting security updates.
It also enabled the option to run the latest build from the master through the `edge` channel.

At last, it would also allow running `ddclient` on the embedded systems running Ubuntu Core, as those systems only support snap packaging.

If merged, an automated build on git commit can be set up at https://snapcraft.io/build with few clicks...

### DIFF
--- a/build-aux/snap/snapcraft.yaml
+++ b/build-aux/snap/snapcraft.yaml
@@ -42,7 +42,7 @@ apps:
 parts:
   ddclient:
     plugin: autotools
-    source: https://github.com/ddclient/ddclient.git
+    source: .
     autotools-configure-parameters:
       - --prefix=/usr
       - --sysconfdir=/etc/ddclient
--- a/build-aux/snap/snapcraft.yaml
+++ b/build-aux/snap/snapcraft.yaml
@@ -1,0 +1,92 @@
+name: ddclient
+summary: Ddclient updates dynamic DNS entries on a wide range of dynamic DNS services
+description: |
+  ddclient is a Perl client used to update dynamic DNS entries
+  for accounts on many dynamic DNS services. It uses curl for internet access.
+  Full list of the supported services can be found at the project page:
+  https://github.com/ddclient/ddclient?tab=readme-ov-file
+
+  Once installed, configure config file and enable service with
+  `$ snap start --enable ddclient.daemon`
+  For more information run `$ ddclient`
+
+type: app
+base: core24
+adopt-info: ddclient
+license: GPL-2.0
+
+platforms:
+  amd64:
+  armhf:
+  arm64:
+
+confinement: strict
+grade: stable
+
+environment:
+  LD_LIBRARY_PATH: ${SNAP}/usr/lib/${CRAFT_ARCH_TRIPLET_BUILD_FOR}
+  PATH: ${SNAP}/usr/bin:/usr/bin
+  PERL5LIB: ${SNAP}/usr/share/perl:${SNAP}/usr/share/perl5:${SNAP}/usr/share/perl-openssl-defaults:${SNAP}/usr/lib/${CRAFT_ARCH_TRIPLET_BUILD_FOR}/perl
+
+apps:
+  ddclient:
+    command: usr/bin/ddclient-wrapper --help
+
+  daemon:
+    command: usr/bin/ddclient-wrapper
+    install-mode: disable
+    daemon: simple
+    plugs:
+      - network
+
+parts:
+  ddclient:
+    plugin: autotools
+    source: https://github.com/ddclient/ddclient.git
+    autotools-configure-parameters:
+      - --prefix=/usr
+      - --sysconfdir=/etc/ddclient
+      - --localstatedir=/var
+    override-pull: |
+      craftctl default
+      stable_tag=$(git tag --list | sort --version-sort | tail -1 | cut -c 2-)
+      # collect versions in the store
+      stable_version="$(snap info ${CRAFT_PROJECT_NAME} | awk '$1 == "latest/stable:" {gsub(/--/,"",$2); print $2 }' || true)" 2>/dev/null
+      beta_version="$(snap info ${CRAFT_PROJECT_NAME} | awk '$1 == "latest/beta:" {gsub(/--/,"",$2); print $2 }' || true)" 2>/dev/null
+      # If there is no stable version, build latest stable_tag
+      # If stable_tag is newer than stable_version, build stable_tag
+      # If beta channel is closed (string '^'), build stable_tag
+      # if we have no version defined, we are building tip from master
+      if [ -z "${stable_version}" ] \
+         || [ "^" = "${beta_version}" ] \
+         || $(dpkg --compare-versions "${beta_version}" "lt" "${stable_tag}" ); then
+        version="${stable_tag}"
+        echo "building tagged version ${version}"
+      else
+        echo "building from the tip..."
+      fi
+      if [ -n "${version-}" ]; then
+        git checkout v${version}
+      else
+        version="$(git describe --tags | cut -c 2-)"
+      fi
+      craftctl set version="${version}"
+    override-build: |
+      craftctl default
+      # disable mail operations
+      sed -i \
+          -e 's/mail=\(.*\)/# mail=\1/g' \
+          -e 's/mail-failure=\(.*\)/# mail-failure=\1/g' \
+          ${CRAFT_PART_INSTALL}/etc/ddclient/ddclient.conf
+      sed -i \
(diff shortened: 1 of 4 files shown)


build-aux/snap/snapcraft.yaml
```diff
@@ -0,0 +1,92 @@
...
+      else
+        version="$(git describe --tags | cut -c 2-)"
+      fi
+      craftctl set version="${version}"
+    override-build: |
+      craftctl default
+      # disable mail operations
+      sed -i \
+          -e 's/mail=\(.*\)/# mail=\1/g' \
+          -e 's/mail-failure=\(.*\)/# mail-failure=\1/g' \
```
rhansen, line 79:
```suggestion
          -e 's/^\(mail=.*\)/# \1/g' \
          -e 's/^\(mail-failure=.*\)/# \1/g' \
```

build-aux/snap/snapcraft.yaml
```diff
@@ -0,0 +1,92 @@
...
+          ${CRAFT_PART_INSTALL}/etc/ddclient/ddclient.conf
```
rhansen, line 80:
```suggestion
          "${CRAFT_PART_INSTALL}"/etc/ddclient/ddclient.conf
```

build-aux/snap/snapcraft.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - network
+
+parts:
+  ddclient:
+    plugin: autotools
+    source: .
+    autotools-configure-parameters:
+      - --prefix=/usr
+      - --sysconfdir=/etc/ddclient
+      - --localstatedir=/var
```
rhansen, line 49: I would expect the autotools plugin to pass appropriate defaults for these parameters.  Does it not?

build-aux/snap/snapcraft.yaml
```diff
@@ -0,0 +1,92 @@
...
+      if [ -z "${stable_version}" ] \
+         || [ "^" = "${beta_version}" ] \
+         || $(dpkg --compare-versions "${beta_version}" "lt" "${stable_tag}" ); then
+        version="${stable_tag}"
+        echo "building tagged version ${version}"
+      else
+        echo "building from the tip..."
+      fi
+      if [ -n "${version-}" ]; then
+        git checkout v${version}
```
rhansen, line 69:
```suggestion
        git checkout "v${version}"
```

build-aux/snap/snapcraft.yaml
```diff
@@ -0,0 +1,92 @@
...
+      # disable mail operations
+      sed -i \
+          -e 's/mail=\(.*\)/# mail=\1/g' \
+          -e 's/mail-failure=\(.*\)/# mail-failure=\1/g' \
+          ${CRAFT_PART_INSTALL}/etc/ddclient/ddclient.conf
+      sed -i \
+          -e 's|'"${CRAFT_PART_INSTALL}"'|/snap/'"${CRAFT_PROJECT_NAME}"'/current|g' \
+          ${CRAFT_PART_INSTALL}/usr/bin/ddclient
+      install -m 755 ${CRAFT_PROJECT_DIR}/ddclient-snap-wrapper \
+                     ${CRAFT_PART_INSTALL}/usr/bin/ddclient-wrapper
```
rhansen, line 85:
```suggestion
          "${CRAFT_PART_INSTALL}"/usr/bin/ddclient
      install -m 755 "${CRAFT_PROJECT_DIR}"/ddclient-snap-wrapper \
                     "${CRAFT_PART_INSTALL}"/usr/bin/ddclient-wrapper
```

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
rhansen, line 1:
```suggestion
#!/bin/sh
```

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
...
+# handle --help as special case
+
+config_file="${SNAP_COMMON}/etc/ddclient/ddclient.conf"
+
+print_help() {
+  echo -e "${SNAP_NAME} uses config file as way to configuration."
+  echo -e "Edit ${config_file} with desired configuration."
+  echo -e "Once config is ready, enable service with:"
+  echo -e "\t$ snap start --enable ${SNAP_NAME}.daemon"
+}
```
rhansen, line 12:
```suggestion
print_help() {
  cat <<EOF
${SNAP_NAME} uses config file as way to configuration.
Edit ${config_file} with desired configuration.
Once config is ready, enable service with:
	$ snap start --enable ${SNAP_NAME}.daemon
EOF
}
```

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
...
+
+if [ "${1}" = "--help" ]; then
+  print_help
+else
+  mkdir -p -m 700 /tmp/cache
+  "${SNAP}"/usr/bin/ddclient \
```
rhansen, line 18:
```suggestion
  exec "${SNAP}"/usr/bin/ddclient \
```

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
...
+           -file "${config_file}" \
+           -cache /tmp/cache/ddclient.cache \
+           -pid /tmp/ddclient.pid \
```
rhansen, line 21:
```suggestion
           --file="${config_file}" \
           --cache=/tmp/cache/ddclient.cache \
           --pid=/tmp/ddclient.pid \
```

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
...
+  echo -e "Once config is ready, enable service with:"
+  echo -e "\t$ snap start --enable ${SNAP_NAME}.daemon"
+}
+
+if [ "${1}" = "--help" ]; then
+  print_help
+else
+  mkdir -p -m 700 /tmp/cache
+  "${SNAP}"/usr/bin/ddclient \
+           -file "${config_file}" \
```
rhansen, line 19: Passing the proper value for `./configure --sysconfdir` should make this unnecessary.

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
...
+           -cache /tmp/cache/ddclient.cache \
```
rhansen, line 20: "Cache" is a misnomer that we are unfortunately stuck with for backwards compatibility.  This really belongs under the directory used for `./configure --localstatedir`, so it shouldn't be necessary to pass `--cache`.

ddclient-snap-wrapper
rhansen, line 21: Does snap need this?  If not, I recommend removing this.

ddclient-snap-wrapper
```diff
@@ -0,0 +1,24 @@
...
+           -pid /tmp/ddclient.pid \
+           --foreground
+           ${@}
```
rhansen, line 23:
```suggestion
           "${@}"
```

rhansen (changes requested):
Thank you very much for this contribution!  I'm not familiar with building snap packages, but I'm very interested in this.  I expect this to benefit lots of users.

beadon:
Apologies for the long delay on this — thank you for the contribution, @kubiko !

I am stepping up here to see if there is a true use case here for merging things since the long quiet after feedback was provided by @rhansen .

I see a couple reasons why ddclient may be a bad fit for snap packages :

1. Snap is a poor fit for a system daemon. ddclient runs as a privileged daemon and reads credentials from /etc/ddclient/ddclient.conf. I agree that for the majority of its functions ddclient does not need to run as priveleged, it only does so because logins details are stored in a config file with filesystem-level access controls.  Snap's filesystem isolation requires the install hook workaround already present in this PR, and @rhansen's concern about config being overwritten on upgrade is a real user-facing risk we'd need to get right. The ```snap_daemon``` user is also deprecated upstream, which adds ongoing maintenance burden.

2. The target use case is narrow. The main benefit cited is running on Ubuntu Core (embedded systems). That's a legitimate niche, but it would drive significant complexity into our build and release process for a small fraction of users. Users on older distros without recent packages may be better served by installing from a tarball.

3. Snap's ecosystem trajectory. Since this PR was opened, community preference has continued shifting toward Flatpak for universal packaging. Snap's Snap Store dependency means a new build/release infrastructure dependency on Canonical's closed platform, which we'd prefer to avoid for a cross-distro tool.

I do see the package maintenance headache and it is *real*, users are not getting the latest software because package maintainers cannot package fast enough and are overwhelmed at the speed in which things are releasing, so auto-packaging solutions are being built to try and address this.

I'll give this a week(?) and then close the PR as a it can always be reopened.

PR #754, opened 2024-10-25
Correct NearlyFreeSpeech.NET example

Zorks:
extra spaces in the protocol and a trailing comma caused connection issues if not corrected.


indrajitr (approved):

indrajitr:
Thanks @Zorks!

PR #766, opened 2024-11-26
Enable arguments on cmdv4 and cmdv6

### LINKED ISSUE #736: cmdv4-args and cmdv6-args options
While trying to resolve https://github.com/ddclient/ddclient/issues/635 which I was running into (my Dynu IP was incorrectly getting set to 1.0.1.1), I tried adding this to my ddclient.conf file:

```
usev4=cmdv4, cmdv4='curl -fs http://ipv4.icanhazip.com/'
```

But ddclient tried to interpret that whole thing as a command: I got the error `sh: line 1: curl -fs http://ipv4.icanhazip.com/: No such file or directory`. I haven't found any documented way to pass arguments to the command specified in the cmdv4/cmdv6 options.

Obviously I can put together a one-liner get-ipv4.sh shell script that calls curl with appropriate arguments. But it would be nice to be able to specify a command *and arguments* via the cmdv4 and cmdv6 options, and not have to use the indirection of a shell script.

Would it be feasible to add `cmdv4-args` and `cmdv6-args` options to ddclient? Or is "just throw together a wrapper shell script" the intended resolution, similar to how `cmd-skip` was deprecated in favor of doing the parsing/filtering in a wrapper script?

greenius:
Do not call `quotemeta` on `cmdv4` and `cmdv6` arguments as it prevents executing commands with arguments.  eg. it is now possible to get an IP address with:

```
usev4=cmdv4
cmdv4="dig +short myip.opendns.com @resolver1.opendns.com"
```

Fixes #736

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,17 @@ repository history](https://github.com/ddclient/ddclient/commits/main).
 
 ## v4.0.0-rc.3 (unreleased work-in-progress)
 
+### Breaking changes
+
+  * The string argument to `--cmdv4` or `--cmdv6` is now executed as-is by the
+    system's shell, matching the behavior of the deprecated `--cmd` option.
+    This makes it possible to pass command-line arguments, which reduces the
+    need for a custom wrapper script.  Beware that the string is also subject to
+    the shell's command substitution, quote handling, variable expansion, field
+    splitting, etc., so you may need to add extra escaping to ensure that any
+    special characters are preserved literally.
+    [#766](https://github.com/ddclient/ddclient/pull/766)
+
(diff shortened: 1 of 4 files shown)


indrajitr:
This might include cases that require more testing.
For example, scenarios like `cmdv4="my\ alt\ path/bin/helper detect_ip"`

rhansen (approved):
@greenius I rebased your commit onto latest `main` and pushed a fixup to refine your commit's message and eliminate an unnecessary variable.  Would you mind reviewing it and letting me know if it looks OK to you?  If so, I'll autosquash the two commits and merge.

greenius:
Amendment looks good to me.

PR #776, opened 2024-12-23
Replace tilde (`~`) with hyphen-minus (`-`) in Git version tags

rhansen:
@indrajitr Would you sanity check this for me?  I went to create a `v4.0.0~rc1` tag and realized that `~` is not permitted in tags.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -48,9 +48,17 @@ use Sys::Hostname;
 #     1.2.3rN           v1.2.3.999.N    1 <= N < 1000; for re-releases, if necessary (rare)
 #
 # A tilde is used to separate "alpha", "beta", and "rc" from the version numbers because it has
-# special meaning for the version comparison algorithms in RPM and Debian:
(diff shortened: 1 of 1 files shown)


indrajitr:
Indeed, we'll have to have `v4.0.0-rc1` and on Debian side mangle `-rc` to `~rc` if necessary. 

This is something [systemd does](https://salsa.debian.org/systemd-team/systemd/-/commit/db2dbed693ac75c88ea6ed923537d18d30fc1cdf) in `debian/watch` to account for it. Similar mangling can be done on Debian side if necessary, but as upstream, yes we would probably want to follow the [semver](https://semver.org/) convention of `-rc1` instead of `~rc1`.

indrajitr (approved):

rhansen:
Given that we can transform the version in Debian (and probably other downstreams), maybe semver is better than `~`.  I'll close this and open a different PR to switch to semver instead.

PR #777, opened 2024-12-24
Use semver 2.0.0 as the human-readable version string format

rhansen:
This avoids the need to escape tilde in tag names.

@indrajitr Mind reviewing this?


indrajitr (approved):
Looks great!

PR #780, opened 2024-12-25
Testing disabled linear PR check—do not merge

rhansen:
this is a test of the new linear PR workflow.  Do not merge.


rhansen:
yay, it works.  Adding the pr-permit-nonlinear label does not prevent the PR from being merged even though the Linear History check is marked as required but skipped.

PR #781, opened 2024-12-27
systemd: Add systemd service settings for better security

indrajitr:
This adds additional settings to improve the security of 'ddclient.service'. 

The settings are based on basic [systemd guidelines][1] and other security-related [guide][2] and [roadmap][3].

[1]: https://www.freedesktop.org/software/systemd/man/systemd.exec.html
[2]: https://www.redhat.com/sysadmin/systemd-secure-services
[3]: https://wiki.debian.org/ReleaseGoals/SystemdAnalyzeSecurity

### DIFF
--- a/sample-etc_systemd.service
+++ b/sample-etc_systemd.service
@@ -9,5 +9,31 @@ Environment=daemon_interval=5m
 ExecStart=/usr/bin/ddclient --daemon ${daemon_interval} --foreground
 Restart=on-failure
 
+SystemCallArchitectures=native
+SystemCallFilter=@system-service
+SystemCallFilter=~ @privileged @resources
+
+CapabilityBoundingSet=
+NoNewPrivileges=yes
+
+ProtectControlGroups=yes
+ProtectKernelTunables=yes
+ProtectKernelModules=yes
+ProtectKernelLogs=yes
+ProtectProc=invisible
+ProtectClock=yes
+ProtectHostname=yes
+
+ProtectSystem=yes
+ProtectHome=yes
+PrivateTmp=yes
+PrivateDevices=yes
+MemoryDenyWriteExecute=true
+
+RestrictRealtime=yes
+RestrictSUIDSGID=yes
+RestrictNamespaces=yes
+RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6
+
 [Install]
 WantedBy=multi-user.target


rhansen (commented):
Are any downstreams running with this?  I'm hesitant to merge this until after 4.0.0 is released because I don't want to break anything while preparing for the release..

Also, please add an entry to `ChangeLog.md`.

indrajitr:
> Are any downstreams running with this?

Most don't (Debian, Fedora/RedHat maintain their own copies) but Arch Linux [does](https://gitlab.archlinux.org/archlinux/packaging/packages/ddclient/-/blob/main/PKGBUILD?ref_type=heads#L49). As the name indicates, this file is expected to be a suggested example, generally used as a reference for the distributions - and adapted based on distribution-specific uniqueness.

> I'm hesitant to merge this until after 4.0.0 is released because I don't want to break anything while preparing for the release..

Indeed! This is not critical and should wait till 4.0.0 is released.

> Also, please add an entry to ChangeLog.md.

Yes, I'll create a separate 4.0.x section and put it in there. We can get back to this after 4.0.0 release.

rhansen:
> > Are any downstreams running with this?
> 
> Most don't (Debian, Fedora/RedHat maintain their own copies) but Arch Linux [does](https://gitlab.archlinux.org/archlinux/packaging/packages/ddclient/-/blob/main/PKGBUILD?ref_type=heads#L49).

I meant:  Do any downstreams currently use the changes in this PR?  Or were these permissions copied from anywhere?  I want to know if these changes have been battle tested.

beadon:
Thanks for this PR — the hardening is well-targeted for a long-running network daemon, and 4.0.0 has now shipped so the timing hold from the earlier discussion is resolved.

A few things worth addressing before merging:

**`RestrictAddressFamilies` needs `AF_NETLINK`**

The proposed list (`AF_UNIX AF_INET AF_INET6`) will silently break users who configure IP detection via `use=if` or `use=fw`-variants. When those modes are active, ddclient spawns `ip -4/-6 -o addr show dev <interface>` as a child process (see `get_ip_from_interface()`). The `ip` command opens an AF_NETLINK socket to query the kernel's routing/address tables — that `socket()` syscall will return `EPERM` under a seccomp filter that doesn't include AF_NETLINK, and the restriction is inherited by child processes. Suggested fix:

```ini
RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6 AF_NETLINK
```

**`MemoryDenyWriteExecute=true`**

This has historically caused surprises with Perl on some platforms when XS modules perform shared-library relocation using temporarily writable+executable mappings. Current ddclient's optional module usage (`JSON::PP` is pure-Perl, `Digest::SHA` has a well-behaved XS backend) is likely fine, but it is worth running a quick smoke-test against the oldest Perl in the CI matrix (Perl 5.30 / Debian oldstable) with this setting active before merging. Also a minor style note: all other settings in the patch use `yes`; this one uses `true`. Both are accepted by systemd, but consistency with the rest of the file reads better as `MemoryDenyWriteExecute=yes`.

**ChangeLog entry**

One was agreed upon earlier in the thread — still needed.

**Optional: section comments**

Since this is a sample file that distributions adapt rather than install verbatim, brief inline comments grouping the directives by purpose (kernel protection, filesystem isolation, syscall filtering, etc.) would help sysadmins understand what they're enabling when they customise it for their package. This is purely cosmetic and not a blocker.

With `AF_NETLINK` added and a ChangeLog entry, this looks ready to go.

PR #785, opened 2025-01-02
Add support for myaddr.tools

### LINKED ISSUE #768: Support for https://myaddr.tools/
Add support for another dyndns service

IP for the domain-name can be updated with a GET / POST / PUT on https://ipv4.myaddr.tools/update

e.g.
wget -O- -t 1 -T 60 --post-data "key=${MYKEY}&ip=self" https://ipv4.myaddr.tools/update

using the provided key -> MYKEY

brianshea2:
Adds support for new dynamic dns service: [myaddr.tools](https://myaddr.tools)
Fixes: #768
Cheers!

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,11 @@ repository history](https://github.com/ddclient/ddclient/commits/main).
     special characters are preserved literally.
     [#766](https://github.com/ddclient/ddclient/pull/766)
 
+### New features
+
+  * `myaddr`: New `protocol` option for updating [myaddr](https://myaddr.tools/)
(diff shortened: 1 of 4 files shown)


brianshea2:
Should I be adding a ChangeLog entry? I'm unsure where in that file I would do so since 4.0.0-rc.1 was just released.

rhansen:
> Should I be adding a ChangeLog entry?

Yes please.

> I'm unsure where in that file I would do so since 4.0.0-rc.1 was just released.

Doesn't really matter—when we rebase onto `main` before merging we'll move it to the right spot if it isn't already there.

I'll review this once v4.0.0 is out the door.  It'll go in the v4.1.0 release.

Thanks for the contribution!

PR #789, opened 2025-01-05
Change default location of `ddclient.conf` to `${sysconfdir}/ddclient`

### LINKED ISSUE #788: Look for `ddclient.conf` in `ddclient` subdirectory by default
`README.md` recommends passing `--sysconfdir=/etc/ddclient` but packagers usually default to `--sysconfdir=/etc`, resulting in confusion like #756.  ddclient should instead look for `ddclient.conf` in `$(sysconfdir)/ddclient` by default.

Alternatively, the documentation could be changed to recommend `--sysconfdir=/etc` and cite `/etc/ddclient.conf` in examples.  However, a dedicated `ddclient` subdirectory is convenient for users with helper scripts (e.g., `cmdv4`).

This will break backwards compatibility, so it will require a major version bump.

rhansen:
fixes #788

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,31 @@ repository history](https://github.com/ddclient/ddclient/commits/main).
 
 ## v4.0.0-rc.2 (unreleased work-in-progress)
 
+### Breaking changes
+
+  * ddclient now looks for `ddclient.conf` in `${sysconfdir}/ddclient` by
+    default instead of `${sysconfdir}`.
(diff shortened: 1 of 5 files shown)


indrajitr:
Can this be considered for 4.0.0? We have quite a few breaking changes for 4.0.0 anyway and waiting for 5.0.0 essentially is kicking the can even further down the road :)

Besides, package maintainers still have a way to keep the legacy path if they choose to. And even for end users it is just a `--file` option.

indrajitr (approved):
Rebased with main as well :)

PR #792, opened 2025-01-08
tests: Fix t/ssl-validate.pl in minimal test environments

### LINKED ISSUE #791: Broken `t/ssl-validate.pl` tests on Debian trixie/sid?
While working on updating the Debian package to v4.0.0-rc.1, `pdebuild` with a Debian trixie (testing) chroot would fail due to failing `t/ssl-validate.pl` tests.  The odd thing is the tests don't fail when run directly on the host (outside the `pbuilder` chroot), even on a system running Debian trixie.  They also don't fail on [salsa](https://salsa.debian.org/debian/ddclient) CI.  There's something about the `pbuilder` chroot environment that is causing the failure, and I haven't been able to root cause it yet.  I'd like to figure this out before releasing v4.0.0 in case there's a recent change to curl/OpenSSL/GnuTLS/etc. that breaks ddclient.

The problem arises after the first `web-ssl-validate=yes` case for each test server (IPv4 and IPv6).  After each of those test cases, the corresponding test web server stops accepting new connections.

```
curl: (7) Failed to connect to 127.0.0.1 port 36147 after 0 ms: Could not connect to server
...
curl: (7) Failed to connect to ::1 port 45111 after 0 ms: Could not connect to server
```

Troubleshooting notes:
  * This is reproducible on two different systems, one running Debian trixie and the other running Ubuntu jammy.
  * Also fails with a sid (unstable) chroot.
  * The test server uses blocking IO and does not fork or spawn a thread to handle an incoming connection, so if `accept` or the request handler hangs then future connection attempts will fail.
  * The test server does TLS negotiation during its `accept` method.
  * The test server's connection backlog is set to 5 (IIUC), so if `accept` or the request handler hangs then I would expect future connection attempts to hang and eventually time out, not be immediately rejected.
  * The connection attempts are failing immediately, not timing out.
  * `kill(0, $server_pid)` succeeds after the `web-ssl-validate=yes` test, so the server process is still running.
  * Using Net::Ping to connect to the server's TCP port fails immediately.
  * After adding some debug log messages, it looks like the test server hangs in `accept`.
  * Maybe `curl` or the TLS library it uses forks a short-lived daemon that maintains a TCP or TLS connection cache between invocations?  Seems unlikely...
  * This isn't a HSTS failure; we're not passing `--hsts` to `curl` and the test server doesn't reply with the `Strict-Transport-Security` header.
  * Passing `--no-keepalive` to `curl` has no effect.
  * Passing `--no-sessionid` to `curl` has no effect.
  * Setting `$IO::Socket::SSL::DEBUG = 3` wasn't very useful:
      * on host (not in chroot):

        ```
        DEBUG: .../IO/Socket/SSL.pm:988: accept created normal socket HTTP::Daemon::ClientConn::SSL=GLOB(0x5635d22cff60)
        DEBUG: .../IO/Socket/SSL.pm:1016: starting sslifying
        DEBUG: .../IO/Socket/SSL.pm:1066: Net::SSLeay::accept -> -1
        DEBUG: .../IO/Socket/SSL.pm:1066: Net::SSLeay::accept -> -1
        DEBUG: .../IO/Socket/SSL.pm:1066: Net::SSLeay::accept -> 1
        DEBUG: .../IO/Socket/SSL.pm:1114: handshake done, socket ready
        DEBUG: .../IO/Socket/SSL.pm:1004: accept_SSL ok
        DEBUG: .../IO/Socket/SSL.pm:1193: local error: SSL read error
        DEBUG: .../IO/Socket/SSL.pm:1516: local error: SSL shutdown error (5)
        DEBUG: .../IO/Socket/SSL.pm:986: no socket yet
        ```

    * in chroot:

        ```
        DEBUG: .../IO/Socket/SSL.pm:988: accept created normal socket HTTP::Daemon::ClientConn::SSL=GLOB(0x5830f14124c0)
        DEBUG: .../IO/Socket/SSL.pm:1016: starting sslifying
        DEBUG: .../IO/Socket/SSL.pm:1066: Net::SSLeay::accept -> -1
        ```

Things to try:
  * [ ] run wireshark/tcpdump to inspect the number of connection attempts and TCP RST packets
  * [ ] separate TCP accept from TLS negotiation
  * [ ] fork between TCP accept and TLS negotiation

rhansen:
fixes #791

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -156,6 +156,7 @@ EXTRA_DIST += $(handwritten_tests) \
 	t/lib/ddclient/Test/Fake/HTTPD/dummy-ca-cert.pem \
 	t/lib/ddclient/Test/Fake/HTTPD/dummy-server-cert.pem \
 	t/lib/ddclient/Test/Fake/HTTPD/dummy-server-key.pem \
+	t/lib/ddclient/Test/Fake/HTTPD/other-ca-cert.pem \
 	t/lib/ddclient/t.pm \
 	t/lib/ddclient/t/HTTPD.pm \
 	t/lib/ddclient/t/ip.pm \
--- a/t/lib/ddclient/Test/Fake/HTTPD/other-ca-cert.pem
+++ b/t/lib/ddclient/Test/Fake/HTTPD/other-ca-cert.pem
@@ -1,0 +1,80 @@
+Certificate:
+    Data:
+        Version: 3 (0x2)
+        Serial Number:
+            6c:bf:34:52:19:4d:c9:29:2b:a6:8b:41:59:aa:c6:c5:1f:a2:bb:10
+        Signature Algorithm: sha256WithRSAEncryption
+        Issuer: CN=Root Certification Authority
+        Validity
+            Not Before: Jan  8 08:24:32 2025 GMT
+            Not After : Jan  9 08:24:32 2125 GMT
+        Subject: CN=Root Certification Authority
+        Subject Public Key Info:
+            Public Key Algorithm: rsaEncryption
+                Public-Key: (2048 bit)
+                Modulus:
+                    00:c3:3d:19:6b:72:0a:9e:87:c0:28:a1:ff:d0:08:
+                    21:55:52:71:92:f2:98:36:75:fc:95:b4:0c:5e:c9:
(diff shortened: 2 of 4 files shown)


indrajitr (approved):
Awesome

Issue #799, opened 2025-01-12
[3.11.2 with OVH] found neither IPv4 nor IPv6 address:


alexisfrjp:
I use pikvm to update my domain for OVH Dynhost.

ddclient version 3.11.2

I use the example in ddclient.conf.
```
# OVH
protocol=ovh
login=mydomain.com-sub
password='XXX'
sub.mydomain.com
```

```
=== opt ====
opt{cache}                           : <undefined>
opt{cmd}                             : <undefined>
opt{cmd-skip}                        : <undefined>
opt{cmdv4}                           : <undefined>
opt{cmdv6}                           : <undefined>
opt{daemon}                          : <undefined>
opt{debug}                           : 1
opt{exec}                            : <undefined>
opt{facility}                        : <undefined>
opt{file}                            : <undefined>
opt{force}                           : <undefined>
opt{foreground}                      : 1
opt{fw}                              : <undefined>
opt{fw-banlocal}                     : CODE(0x7a58a0)
opt{fw-login}                        : <undefined>
opt{fw-password}                     : <undefined>
opt{fw-skip}                         : <undefined>
opt{fw-ssl-validate}                 : <undefined>
opt{fwv4}                            : <undefined>
opt{fwv4-skip}                       : <undefined>
opt{fwv6}                            : <undefined>
opt{fwv6-skip}                       : <undefined>
opt{geturl}                          : <undefined>
opt{help}                            : <undefined>
opt{host}                            : <undefined>
opt{if}                              : <undefined>
opt{if-skip}                         : CODE(0x7a58a0)
opt{ifv4}                            : <undefined>
opt{ifv6}                            : <undefined>
opt{ip}                              : <undefined>
opt{ipv4}                            : <undefined>
opt{ipv6}                            : <undefined>
opt{list-devices}                    : CODE(0x88fde0)
opt{list-protocols}                  : CODE(0x8aaee0)
opt{list-web-services}               : CODE(0x8c7a30)
opt{login}                           : <redacted>
opt{mail}                            : <undefined>
opt{mail-failure}                    : <undefined>
opt{max-interval}                    : 2160000
opt{max-warn}                        : <undefined>
opt{min-error-interval}              : 300
opt{min-interval}                    : 30
opt{options}                         : <undefined>
opt{password}                        : <redacted>
opt{pid}                             : <undefined>
opt{postscript}                      : <undefined>
opt{priority}                        : <undefined>
opt{protocol}                        : <undefined>
[...]

PR #800, opened 2025-01-15
Update ddclient.conf.in, fixing OVH

alexisfrjp:
Adding `use=web` for `protocol=ovh` to work.

#799


rhansen:
This line doesn't belong here because `use=web` is not specific to `protocol=ovh`.  (The method used to discover the host's IP address is unrelated to how a service is updated with the discovered IP address.)

alexisfrjp:
Yes, sorry about that, thanks!

PR #810, opened 2025-02-05
Update ddclient.in to determine perl path using env call

vrtareg:
Using env to determine perl path as it is not usually in /usr/bin for all systems. For FreeBSD for example it is under /usr/bin/local

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 ######################################################################
 #
 # DDCLIENT - a Perl client for updating DynDNS information


StantonMatt:
Thanks, this change makes sense to me. `configure.ac` runs `./ddclient.in --version=short` before the configure script has found and substituted `PERL`, so using `#!/usr/bin/env perl` avoids assuming Perl is installed at `/usr/bin/perl`.

The generated `ddclient` script should still use the configured Perl path. `Makefile.am` rewrites the first-line shebang matching `^#!.*perl$` to `#!$(PERL)`, and that pattern still matches `#!/usr/bin/env perl`.

One small note: the FreeBSD path in the PR description looks like a typo; Perl is typically under `/usr/local/bin/perl`, not `/usr/bin/local`. The branch is also quite far behind `main`, but the effective patch is still just this one-line shebang change.

PR #811, opened 2025-02-09
Zero index sh args in t/geturl_respones.pl

Gnarwhal:
This test currently breaks for me because the version of `sh` I use (a fork of FreeBSD's `sh`) does not treat `--` as an argument (rather than treating it as $0). Removing the `--` and using `$0` seems to be more portable. In addition to my default `sh` I have tried this with bash, dash, and zsh which all seem to work.

### DIFF
--- a/t/geturl_response.pl
+++ b/t/geturl_response.pl
@@ -5,7 +5,10 @@ eval { require 'ddclient'; } or BAIL_OUT($@);
 # Fake curl.  Use the printf utility, which can process escapes.  This allows Perl to drive the fake
 # curl with plain ASCII and get arbitrary bytes back, avoiding problems caused by any encoding that
 # might be done by Perl (e.g., "use open ':encoding(UTF-8)';").
-my @fakecurl = ('sh', '-c', 'printf %b "$1"', '--');
+#
+# Use $0 rather than $1 to avoid relying on -- being treated as argv[0],
+# which some sh variants do not handle correctly.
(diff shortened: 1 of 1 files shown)


beadon:
Thanks for the patch. The change looks correct — `$0` in `sh -c 'cmd' arg0 arg1...` is POSIX-standard for the first extra argument, and removing the `--` placeholder is cleaner.

Could you clarify one thing? In standard POSIX shells, `sh -c 'cmd' '--' arg` should set `$0=--` and `$1=arg`. If your shell variant is not putting `--` into `$0`, that sounds like a departure from POSIX — do you know whether it's treating `--` as an option terminator instead, or doing something else entirely? Just curious about the root cause, it won't change the outcome here.

The fix itself is sound and the `$0` approach works correctly on bash, dash, and zsh. One small request before we merge: could you add a brief note to the existing comment explaining why we use `$0` rather than `$1`? Something like:

```perl
# Use $0 rather than $1 to avoid relying on -- being treated as argv[0],
# which some sh variants do not handle correctly.
my @fakecurl = ('sh', '-c', 'printf %b "$0"');
```

That way future readers won't wonder why the conventional `$1`/`--` pattern was avoided.

Gnarwhal:
Yes I believe that's correct. From the man pages of my 'sh' implementation:
> A “--” or plain ‘-’ will stop option processing and will force the remaining words on the command line to be treated as arguments.

indrajitr (approved):
LGTM!

PR #834, opened 2025-05-29
doc(cfg): cf API token only needs 'Zone - DNS - Edit' perm

litlighilit:
For cloudflare, when using API token,

as I test, only such one perm is added and the update succeeds.

Also, the Zone resources doesn't have to be "Include - All zones";
I just use "Include - Specific zone - ..."

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -5860,7 +5860,7 @@ Configuration variables applicable to the 'cloudflare' protocol are:
   protocol=cloudflare          ##
   server=fqdn.of.service       ## defaults to api.cloudflare.com/client/v4
   login=service-login          ## login email when using a global API key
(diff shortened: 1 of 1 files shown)


beadon:
Apologies for the slow review.

The `Zone:Zone:Read` permission is actually required — the Cloudflare protocol queries `/zones/` to look up the zone ID by name before it can update any DNS records. Without `Zone:Zone:Read`, that lookup will fail with a 403 even if `Zone:DNS:Edit` is present. It may have appeared to work if a previous token or session had already cached the zone ID, or if there is another permission in your token granting implicit zone listing access.

The corrected minimum set for API token auth remains:
- `Zone:DNS:Edit` — to create/update DNS records
- `Zone:Zone:Read` — to look up the zone ID from the zone name

On scope: restricting to a specific zone (rather than "Include - All zones") is perfectly fine and is actually the recommended security practice. That part of your observation is correct and worth preserving.

That said, this section of the docs is likely to be refreshed as part of other in-flight work, so I'd suggest holding off on further changes here for now.

Issue #756, opened 2024-10-31
Where is the config file on debian 12


Smartich0ke:
So I installed ddclient thruough apt on debain 12 but when I test it with `sudo ddclient -daemon=0 -debug -verbose -noquiet` it returns with:
```
WARNING:  skipping host: REDACTED: 'zone=' is an invalid fully qualified host name.
WARNING:  skipping host: *.REDACTED: 'zone=' is an invalid fully qualified host name.
WARNING:  skipping host: www.REDACTED 'zone=' is an invalid fully qualified host name.
WARNING:  skipping host: REDACTED: 'zone=' is an invalid fully qualified host name.
WARNING:  skipping host: *.REDACTED: 'zone=' is an invalid fully qualified host name.
WARNING:  skipping host: www.REDACTED: 'zone=' is an invalid fully qualified host name.

```
Its supposedly running off the config that I set up when it installed through the tui thing and I want to change that because obviously it didnt work. So I go to edit `/etc/ddclient/ddclient.conf` only to find that `/etc/ddclient` doesnt even exist. I put the config in there anyway:
```
daemon=300 # check every 300 seconds
syslog=yes # log update msgs to syslog
pid=/var/run/ddclient.pid # record PID in file
ssl=yes # use TLS
use=web # get IP with website below
web='https://cloudflare.com/cdn-cgi/trace'
web-skip='ip=' # IP address above is after 'ip='


protocol=cloudflare, \
zone=REDACTED, \
ttl=1,
password='REDACTED',

```

But it still appears to be running off some mystery config file somewhere that was created during installation despite all the documentations saying it should read from /etc/ddclient/ddclient/conf.

What's going on here? Where is the actual config?

rhansen:
`README.md` suggests passing `--sysconfdir=/etc/ddclient` to the `configure` script when building ddclient, which is how you would get `/etc/ddclient/ddclient.conf`.  The Debian package passes `--sysconfdir=/etc` instead, which is why it differs.  I'll see if I can get that changed for Debian's v4.0.0 package.

I opened #788 to change ddclient to look in a `ddclient` subdirectory by default.

PR #838, opened 2025-06-27
docs: Fixed build options to match the official documentation.

gretchi:
I have fixed a discrepancy in the reference path for `ddclient.conf` between the **official documentation** and the README.md build default.

Specifically:

- The **README.md** / build-time default expects the config at `/etc/ddclient.conf`.
- The **official documentation** (https://ddclient.net/#configuration) describes the default config file as `/etc/ddclient/ddclient.conf`.

I have submitted a patch to align these and ensure consistency across both usages.

**Related Issue:**

- [#756](https://github.com/ddclient/ddclient/issues/756): Where is the config file on debian 12
- [#788](https://github.com/ddclient/ddclient/issues/788): Look for ddclient.conf in ddclient subdirectory by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ operating system. See the image to the right for a list of distributions with a 
      ```shell
      ./configure \
          --prefix=/usr \
-         --sysconfdir=/etc \
+         --sysconfdir=/etc/ddclient \
          --localstatedir=/var
      make
      make VERBOSE=1 check


beadon:
Apologies for the delay in reviewing this.

The proposed change sets `--sysconfdir=/etc/ddclient`, but `confdir` defaults to `${sysconfdir}/ddclient`, so the config would end up at `/etc/ddclient/ddclient/ddclient.conf` — one level too deep.

The correct build invocation to get `/etc/ddclient/ddclient.conf` is:

```shell
./configure \
    --prefix=/usr \
    --sysconfdir=/etc \
    --localstatedir=/var
```

That said, the user confusion this is trying to address is real and valid. PR #903 adds an "Upgrading from 3.x" section to the README that covers the config path change properly, along with a runtime warning when ddclient starts with an empty config but finds a file at the old 3.x location. Once that merges, it should address the same pain point more completely. I'll close this PR then as superseded.

PR #867, opened 2025-12-25
improved query parameter encoding

elifa:
I had some issues with query strings not being correcly encoded for some characters in hostnames, zones and tokens.

This solved that problem by encoding all query string values.

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -2834,6 +2834,7 @@ sub geturl {
     my $headers = $params{headers} // '';
     my $method = $params{method} // 'GET';
     my $data = $params{data} // '';
+    my $query = $params{query};
 
     my $reply;
     my $server;
@@ -2859,6 +2860,16 @@ sub geturl {
     $server = $url;
     $server =~ s%[?/].*%%;
     $url    =~ s%^[^?/]*/?%%;
+
+    ## append query parameters if provided
+    if (defined($query) && ref($query) eq 'HASH' && %$query) {
+        my $query_string = encode_www_form_urlencoded($query);
+        if ($url =~ /\?/) {
+            $url .= '&' . $query_string;
+        } else {
+            $url .= '?' . $query_string;
+        }
+    }
 
     $protocol = ($use_ssl ? "https" : "http");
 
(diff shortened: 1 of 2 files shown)


beadon (changes requested):
The core idea here is sound — separating query parameter construction from URL building and tightening `encode_www_form_urlencoded` to use a proper unreserved-character allowlist is a genuine improvement. That said, the breadth of the refactor (16+ protocols) introduces several behavioral regressions that need to be resolved before this can merge.

## Must fix

### 1. Njalla flag-style parameters are silently changed

The original code appends `&auto` and `&quiet` as valueless flags:

```perl
$url .= (($auto eq 1)) ? '&auto' : '';
$url .= (($quietreply eq 1)) ? '&quiet' : '';
```

The new code assigns empty-string values:

```perl
$query{auto} = '' if ($auto eq 1);
$query{quiet} = '' if ($quietreply eq 1);
```

`encode_www_form_urlencoded` will emit `auto=` and `quiet=` (with a trailing `=`), not `auto` and `quiet`. Some servers distinguish between `?auto` and `?auto=`. The Njalla API should be verified to accept the latter form.

### 2. FreeDNS debug output now leaks the update token

The original code deliberately redacted the unique update token in debug logs:

```perl
($url_tmpl = $url) =~ s/\?.*\&/?<redacted>&/;
debug("updating: %s", $url_tmpl);
```

The new code logs the raw URL:

```perl
debug("updating: %s", $url);
```

FreeDNS update URLs embed a one-time-use secret token. Logging it in debug mode exposes it in log files. The redaction must be preserved.

## Significant concern

### 3. dyndns2 multi-host commas and IPv6 colons are now percent-encoded

The updated test expectations reveal a behavioral change:

```
- wantquery => 'hostname=h1,h2&myip=192.0.2.1'
+ wantquery => 'hostname=h1%2Ch2&myip=192.0.2.1'

- wantquery => 'hostname=h1&myip=2001:db8::1'
+ wantquery => 'hostname=h1&myip=2001%3Adb8%3A%3A1'
```

The dyndns2 protocol specification explicitly defines `hostname` as a comma-separated list of hosts, and `myip` carries a literal IP address. Conformant servers will URL-decode the query before parsing, so `%2C` should round-trip to `,` correctly. But many dyndns2-compatible services are simple implementations that do not URL-decode query values before splitting — in the field this is likely to break multi-host updates and IPv6 on a range of providers.

The test update masks this regression rather than confirming compatibility. Before merging this should be tested against at least a few of the common dyndns2-compatible services with a multi-host config and an IPv6 address.

## Scope

The PR touches 17 protocol handlers to solve a problem the author observed in (presumably) one or two. Limiting the initial change to the affected protocols and the `encode_www_form_urlencoded` fix would reduce risk considerably. The broader refactoring can follow as a separate PR once the encoding changes are validated.

PR #870, opened 2026-02-12
digitalocean: Save IPs to correct variables

tchebb:
This provider has been using the wrong variable names (`ip-ipv4`/`ip-ipv6` instead of just `ipv4`/`ipv6`) since the beginning, but it was a silent error until v4.0.0, which logs an "ignoring unrecognized recap variable" message thanks to the refactoring from #740. Fix the issue.


indrajitr:
Thanks @tchebb!

PR #876, opened 2026-03-25
Added support for new Hetzner Cloud Api - update

cr3:
Update to https://github.com/ddclient/ddclient/pull/852 which has been inactive since October 12, 2025 (5 months ago). These changes are urgent because they provide a migration path for Hetzner users whose current ddclient setup will stop working on end of May, 2026 (in 2 months).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,13 @@ repository history](https://github.com/ddclient/ddclient/commits/main).
 
 ## v4.0.1-alpha (unreleased work-in-progress)
 
+### Provider updates
+
+  * `hetzner`: Update API endpoint and implementation to align with provider recommendation.
+    [852](https://github.com/ddclient/ddclient/pull/852)
+    [876](https://github.com/ddclient/ddclient/pull/876)
+    Provider notice: https://status.hetzner.com/incident/c2146c42-6dd2-4454-916a-19f07e0e5a44
+
 ### Bug fixes
 
   * `gandi`: Fix %recap tag for ip.
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ Dynamic DNS services currently supported include:
   * [Freemyip](https://freemyip.com)
   * [Gandi](https://gandi.net)
   * [GoDaddy](https://www.godaddy.com)
+  * [Hetzner](https://hetzner.com)
   * [Hurricane Electric](https://dns.he.net)
   * [Infomaniak](https://faq.infomaniak.com/2376)
   * [INWX](https://www.inwx.com/)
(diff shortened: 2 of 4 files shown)


LKaemmerling:
(Hetzner Cloud DNS Lead here) The old DNS API will be shut down on 2026-05-20 as described in the [timeline](https://docs.hetzner.com/networking/dns/faq/beta/#timeline-1).

30350n:
Just gave this a spin and this seems to work fine and resolve the JSON issue I mentioned in #852 :+1:

indrajitr:
I'll take a look in next few days, apologies for the delay in getting back to this.

30350n:
What da slop

cr3:
Fixed some concerns with Copilot, others myself. Enjoy!

cr3:
Gengle nudge to merge this pull request...

indrajitr:
Sorry for the delay in getting back to this @cr3!

As it stands with this PR, we'll have two protocols related to Hetzner (`hetzner` and `hetznercloud`). Given that old Hetzner endpoint is going away, `hetzner` protocol will be non-functional. 

So would it be worth considering updating the existing `hetzner` protocol to support `hetznercloud` instead of creating yet another (seemingly similar) protocol? This just avoids more confusion about the protocol to use.

enspiro:
my DNS configuration was now migrated enforced to the new API. So looks like this is not end of May.
What can i do to make this working again in the OPNsense?

LKaemmerling:
> my DNS configuration was now migrated enforced to the new API. So looks like this is not end of May. What can i do to make this working again in the OPNsense?

Yeah, as lined out in the Migration timeline (which i linked above): https://docs.hetzner.com/networking/dns/faq/beta/#timeline (End of April Migration of all remaining zones, the last shutdown will happen in May)

cr3:
> Sorry for the delay in getting back to this @cr3!
> 
> As it stands with this PR, we'll have two protocols related to Hetzner (`hetzner` and `hetznercloud`). Given that old Hetzner endpoint is going away, `hetzner` protocol will be non-functional.
> 
> So would it be worth considering updating the existing `hetzner` protocol to support `hetznercloud` instead of creating yet another (seemingly similar) protocol? This just avoids more confusion about the protocol to use.

Sure, the last commit now replaces hetzner with the hetznercloud implementation.

cziebuhr:
For me, `hetznercloud` is less confusing. I assume the api token is not migrated on hetzner side, because that would also provide access to other services on hetzner cloud. So a user has to adapt the configuration anyway.

cr3:
> For me, `hetznercloud` is less confusing. I assume the api token is not migrated on hetzner side, because that would also provide access to other services on hetzner cloud. So a user has to adapt the configuration anyway.

I'm easy either way - whatever is most aligned with ddclient. Please let me know if reverting the last commit is needed to merge this branch.

indrajitr (approved):

indrajitr:
Finally! Thanks @lecram89 and @cr3!

enspiro:
by when this change will be available on opnsense release?

PR #878, opened 2026-04-10
fix: add proportional jitter to reduce synchronized API calls

jhutchings1:
Hello, I'm a product manager at @Cloudflare who oversees our API platform. Our logs show that we get millions of clients calling our DDNS endpoints every 5 minutes. This is a noticeable increase over our typical baseline, and contributes to performance degradation. We would like to smooth out the API traffic so that the spikes aren't so prominent, and we have identified your project as one to reach out to. 

I added some code that introduces proportional jitter to the DDNS update interval, which should help reduce synchronized API calls and improve performance for us and all the other services that you're using. 

Let me know if you have any questions on this, happy to take your suggestions on this as well!

PR #888, opened 2026-05-18
feat: Add jitter percentage to the daemon interval

indrajitr:
Add jitter to spread API calls across clients and
reduce overloading upstream DNS provider when many
clients are running with same daemon interval.
 
For example, for a jitter of 60 seconds and a daemon
interval of 10 minutes, a jitter of [0, 60s) would
be applied to the the daemon interval.

This implementation is based on #878 by @jhutchings1.

Changes in this implementation (compared to #878):
- Added global `jitter` configuration defaulting to 20% of `daemon` if `daemon` is set
- Added an external knob to allow configurability (for example local `nsupdate` might be expected to be instantaneous)
- This would also allow changing the jitter (like `daemon`) as CLI param - particularly useful when ddclient is invoked via systemd service or systemd timer (for example using [`RandomizedDelaySec`](https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html#RandomizedDelaySec=)).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,12 @@ repository history](https://github.com/ddclient/ddclient/commits/main).
 
 ## v4.0.1-rc.2 (unreleased work-in-progress)
 
+### New feature
+
+* Add a jitter to the daemon interval when in daemon mode to spread the web service
+  API calls across clients and reduce load on upstream DNS providers when many
+  clients are running with the same daemon interval.
+
 ## 2026-05-16 v4.0.1-rc.1
 
 ### Provider updates
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,11 +65,11 @@ handwritten_tests = \
 	t/group_hosts_by.pl \
 	t/header_ok.pl \
 	t/interval_expired.pl \
-	t/jitter.pl \
 	t/is-and-extract-ipv4.pl \
 	t/is-and-extract-ipv6.pl \
 	t/is-and-extract-ipv6-global.pl \
 	t/logmsg.pl \
+	t/jitter.pl \
 	t/parse_assignments.pl \
 	t/protocol_directnic.pl \
 	t/protocol_dnsexit2.pl \
--- a/Makefile.am
+++ b/Makefile.am
@@ -65,6 +65,7 @@ handwritten_tests = \
 	t/group_hosts_by.pl \
 	t/header_ok.pl \
 	t/interval_expired.pl \
+	t/jitter.pl \
 	t/is-and-extract-ipv4.pl \
 	t/is-and-extract-ipv6.pl \
 	t/is-and-extract-ipv6-global.pl \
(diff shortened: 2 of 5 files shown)


indrajitr:
@jhutchings1, would really appreciate your feedback on this. Thank you!

beadon:
The design is solid and the worked example in both the PR description and the inline comment makes the intent clear. Two issues need to be addressed before merge; one is a runtime bug.

**Blocker: fractional sleep can cause an infinite loop**

`rand($jitter)` returns a float, which makes `$daemon` fractional after the addition (e.g., 347.23 for a 300s daemon with a ~47s jitter sample). The sleep loop then reaches a state where `$left` is less than 1 second (e.g., 0.23). At that point, `sleep 0.23` is silently truncated by Perl to `sleep 0`, which returns 0, so `$left` never decreases and the loop runs forever.

The fix is easy: use `int(rand($jitter))` so the jitter added is always a whole number of seconds, keeping `$daemon` integer throughout.

**Important: the test does not exercise production code**

`t/jitter.pl` has two mismatches with the actual implementation:

The test defines a local `compute_jitter()` using `int(rand($interval * 0.2))` and comments that this is how ddclient computes jitter, but the production code uses `rand($jitter)` with a configurable `$jitter` (not derived inline from the interval), and does not apply `int()`. If the blocker above is fixed with `int()`, the test formula will accidentally match, but the test still won't be testing the right thing.

The `0.2` factor appears as a bare literal in both the formula and the `$max_jitter` bound, treating the configured default (20% of daemon when `jitter` is not explicitly set) as a correctness invariant. Since `jitter` is a user-configurable option, the 20% bound is not a behavioral property to assert: a user who sets `jitter=120` against a 300s daemon would get valid sleep values up to 420s, well outside the range the test considers correct.

The test also stubs `ddclient::sleep`. The daemon loop calls unqualified `sleep`, which resolves to `CORE::sleep`, not the stub. The stub is never invoked by production code, so the test will pass even if the jitter logic is removed entirely.

A more meaningful test would set `opt('daemon')` and `opt('jitter')`, override `CORE::GLOBAL::rand` to return a known value, and then assert that the computed sleep duration falls in the expected range. Alternatively, extracting the jitter application into a small named function would make it directly testable without needing to intercept builtins.

**Nit: default when daemon is not configured**

When `daemon` is not set, `opt('jitter')` resolves to `interval('60s')` = 60 via the code-ref default. This has no runtime effect (jitter is only applied inside `if ($daemon)`), but it means `--help` output shows a 60s jitter default for users not running in daemon mode, which is confusing. Defaulting to `0` when daemon is not configured would be cleaner.

**Nit: Makefile.am ordering**

`t/jitter.pl` is inserted between `interval_expired.pl` and `is-and-extract-ipv4.pl`. Since `j` sorts after `i`, it should follow all `is-*` entries, before `logmsg.pl`.

The rest looks good: the code-ref default pattern is correctly handled by `default()`, the `ne 'inf'` guard correctly matches what `interval()` returns for infinite values, `$daemon` is reset at the top of each loop iteration so jitter does not accumulate, and the CLI option naming is consistent with `--daemon`.

jhutchings1:
Thanks again for jumping on this. PR looks like it should do the job. Thank you!

PR #892, opened 2026-05-20
cloudflare: add support for the 'comment' config variable

### LINKED ISSUE #826: Add "comment" to the list of updatable fields - at least at Cloudflare
The comment field can be updatable at Cloudflare by API.

Here are my DNS settings:
Image
The first three records are maintained by an app with its own updater and you can see the Comment field has been used - the blue speech bubble. The next four are maintained by ddclient and the comment field is not appearing. I have tried using "comment=" in the Cloudflare section but ddclient says it is not a valid field and is ignored.

It would be nice if the comment could be updated as well.

I am using ddclient 3.10.0-2 in Debian 12.

beadon:
## Summary

- Adds a new optional `comment` config variable to the `cloudflare` protocol
- When set, the comment is included in the PATCH request to the Cloudflare DNS API, which supports it natively
- Replaces the hand-interpolated JSON string with `encode_json` for correct escaping

Closes #826.

## Configuration example

```
protocol=cloudflare
zone='example.com'
login='token'
password='your-api-token'
comment='managed by ddclient'
example.com
```

## Test plan

- Tested live against the Cloudflare API: confirmed `comment` field is accepted in a PATCH to `/zones/{zone_id}/dns_records/{record_id}` and returned correctly in the response
- Verified ddclient sends `{"comment":"managed by ddclient","content":"<ip>"}` in the PATCH body when `comment=` is set
- Verified no regression when `comment=` is omitted: PATCH body is `{"content":"<ip>"}` as before

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -937,6 +937,7 @@ our %protocols = (
         'examples'   => \&nic_cloudflare_examples,
         'cfgvars' => {
             %{$cfgvars{'protocol-common-defaults'}},
+            'comment'      => setv(T_STRING, 0, undef,                          undef),
             'login'        => setv(T_LOGIN,  0, 'token',                        undef),
             'min-interval' => setv(T_DELAY,  0, interval('5m'),                 0),
             'server'       => setv(T_FQDNP,  0, 'api.cloudflare.com/client/v4', undef),
@@ -5985,7 +5986,9 @@ sub nic_cloudflare_update {
                 debug("DNS '$type' record ID: $dns_rec_id");
                 # Set domain
                 $url = "https://" . opt('server', $domain) . "/zones/$zone_id/dns_records/$dns_rec_id";
-                my $data = "{\"content\":\"$ip\"}";
+                my %patch = (content => $ip);
+                $patch{comment} = opt('comment', $domain) if defined(opt('comment', $domain));
+                my $data = encode_json(\%patch);
                 $reply = geturl(proxy => opt('proxy'),
                                 url => $url,
                                 headers => $headers,


indrajitr:
To me this looks like a feature creep. I am not sure we want ddclient to start veering into extended features on a per provider basis that aren't supported by the spec in general and only supported by a very small subset of providers.

Let's keep this PR open though. We need to have spec/design discussion around this first.

Meanwhile, one can used `postscript` to apply customization on per host basis.

beadon:
I am confirming if other providers accept comments on their DNS entries like Cloudflare.  So far, it seems to be very unique to Cloudflare as you point out, I thought I had a valid list of 5 other providers, but on further researching these additional providers I see they operate differently.

Since the call to write the IP and the comment occurs in one API call, it seemed best to make the one call instead of use postscript to attach a comment to the DNS entry afterwards.

@indrajitr : Just so I understand your concern:  since this comment variable is only used for Cloudflare implementations the place it's currently located means that too many other providers would have access to it, and this would lead to more mistakes because people would include comments when their providers would not support this feature ?

It's definitely a neat feature I think people would benefit from, and maybe encourage other providers to support it too.

indrajitr:
> I am confirming if other providers accept comments on their DNS entries like Cloudflare. So far, it seems to be very unique to Cloudflare as you point out, I thought I had a valid list of 5 other providers, but on further researching these additional providers I see they operate differently.
> 
> Since the call to write the IP and the comment occurs in one API call, it seemed best to make the one call instead of use postscript to attach a comment to the DNS entry afterwards.
> 
While, I agree `postscript` isn't optimal in this particular case, it does provide a handy general purpose option to apply a _post-processing_ script to perform additional actions.

> @indrajitr : Just so I understand your concern: since this comment variable is only used for Cloudflare implementations the place it's currently located means that too many other providers would have access to it, and this would lead to more mistakes because people would include comments when their providers would not support this feature ?

Main concern is that it does not have anything to do with DNS record. `ddclient` is a tool to update a DNS `A` or `AAAA` record (and in some cases `MX` record) lazily/dynamically pushed from the host. `ddclient` is not a DNS management or IPAM tool as such.

> 
> It's definitely a neat feature I think people would benefit from, and maybe encourage other providers to support it too.

Agreed, and so is the ability to set `private_routing` and `proxied` flags as part of Cloudflare API. But they are probably anything _but_ dynamic. It just so happens that they are exposed throught Cloudflare API. Besides, `comment`, `private_routing`, `proxied` etc. are more likely related to the corresponding hostname and not related to what `A`/`AAAA` record that hostname has.

I personally, use `comment` (for Cloudflare), `note` (for Porkbun) meticulously - but I don't expect them to be dynamically updated based on the IP address change.

PR #897, opened 2026-05-24
Fix: support for EasyDNS HTML resposes

Glandvador:
This fixes operation when EasyDNS responds with HTML messages instead plain text. The change mainly strips the HTML tags for a poor man converter to text, that seems good enough to allow the retrieval of the transaction result by the existing code. Also adds two new result codes 'TOO_FREQ' and 'OK', I suppose used when HTML messages. I don't know about others codes (I only encounter these two), but in any case there is a fallback already in place for unknown codes.
 - Add TOO_FREQ to the errors list (change the text: frequency is high, period is short).
 - Strip HTML tags from the reply: $body =~ s|<.+?>||g;. This does nothing on plain text.
 - Add missing 'OK' to the $resultcode_re. It is checked later, but has no effect without that.
 - Initialise $status to avoid 'Use of uninitialized value ...' warnings.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,9 +7,12 @@ repository history](https://github.com/ddclient/ddclient/commits/main).
 
 ### New feature
 
-* Add a jitter to the daemon interval when in daemon mode to spread the web service
-  API calls across clients and reduce load on upstream DNS providers when many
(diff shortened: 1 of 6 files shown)


indrajitr:
Applied in #911. Thanks @Glandvador!

Issue #609, opened 2023-12-20
Add convenient support for Termux


YuruiHong:
I have succeeded in installing `ddclient` following the advice [here](https://github.com/termux/termux-packages/issues/2987#issuecomment-468948822). However, the process is not perfect as I have to use `sed` to substitute the paths for the compatibility of Termux. Is it possible to check whether the script is running under Termux and provide different paths accordingly, or provide a separate version of archive for Termux users?

PS: The advice itself is actually outdated, since the structure of `ddclient` has changed, and we have to `curl` the source code archive instead.

LenardHess:
You might be able to use the build to change the paths  - see https://github.com/ddclient/ddclient#manual-installation
(I haven't looked deep into those build steps yet)

beadon:
Thanks for raising this. In **ddclient 4.x** the two pain points for Termux are addressed without any Termux-specific code:

### 1. Paths — configurable at build time

All paths that were hardcoded in 3.x are now substituted at `./configure` time.  
For Termux, build with the Termux prefix and you get correct paths everywhere:

```shell
./autogen       # if building from a git clone
./configure --prefix=/data/data/com.termux/files/usr
make
make install
```

This sets the config dir (`/data/data/com.termux/files/usr/etc/ddclient`), the cache dir (`/data/data/com.termux/files/usr/var/cache/ddclient`), and the binary path without needing `sed` or any post-install patching.

### 2. Logging / daemon mode — foreground flag and clean STDERR

In 4.x, all diagnostic output goes to **STDERR** (not mixed with STDOUT). When ddclient forks as a daemon it redirects STDOUT and STDERR to `/dev/null`, which is correct daemon behaviour but unhelpful if you want visible output.

Run with `--foreground` to skip the fork:

```shell
ddclient --foreground
```

Output stays on your terminal via STDERR. `--syslog` is opt-in and off by default, so you don't need a `syslogd` running.

### 3. One-shot / cron-style execution

Without `--daemon`, `ddclient` runs once and exits — works well with Termux's task schedulers (Termux:Tasker, `crond`, etc.):

```shell
ddclient     # update once, exit
```

---

No changes needed to support Termux in 4.x — it should work out of the box once you configure with the right prefix. Closing this. Please reopen if you hit a specific problem on 4.x.

PR #900, opened 2026-05-25
feat: add --run-once flag to update once and exit

### LINKED ISSUE #856: How to run ddclient once in the foreground without sleeping
With 3.x, I had `ddclient -daemon=0 -debug` working and could skip the `DEBUG:   > sleep 60` part (60 is from my config). With 4.x `ddclient --daemon 0 --foreground --debug --verbose` also waits with the config-file based timeout.

BTW did you notice that most of the [example files](https://github.com/ddclient/ddclient/blob/main/sample-etc_ppp_ip-up.local#L39) follow the <4.x switch conventions?

beadon:
## Summary

- Adds a `--run-once` boolean option that overrides any daemon interval (from config or `--daemon` flag), runs a single update cycle, and exits
- Implies `--foreground` (no fork), so the caller gets a synchronous exit code (0 = success, 1 = failure)
- Does not modify `opt('daemon')` itself — only suppresses the loop after the first cycle, so cached state and recap files are still written normally
- Adds `t/run_once.pl` test covering default, override, and no-op cases

## Context

**One-shot is already the default** when no `daemon` interval is set — either in `ddclient.conf` or via `--daemon`. Without a daemon interval, ddclient performs one update cycle and exits. No flag is needed for that case.

`--run-once` exists for the common production scenario where `ddclient.conf` already contains `daemon=<interval>` (e.g. `daemon=5m` to drive a long-running service), but you also need to trigger a single synchronous update from a script, network dispatcher, or `ppp-up` hook — **without modifying the config file**. In that case there is currently no clean way to override the daemon interval: `--daemon=0` is silently clamped to 60 seconds (the enforced minimum), and `--foreground` only suppresses the fork — the sleep loop still runs.

Typical use:

```bash
# one-shot update, foreground implied, exit code reflects success/failure
ddclient --run-once

# with extra visibility
ddclient --run-once --verbose --debug
```

## Motivation

Closes #856.

## Test plan

- [ ] `make check` passes (includes new `t/run_once.pl`)
- [ ] `ddclient --run-once` exits after one cycle with no daemon sleep
- [ ] `ddclient --daemon 300 --run-once` also exits after one cycle (flag overrides daemon interval)
- [ ] `ddclient --run-once` does not fork (foreground implied)
- [ ] `ddclient` without any daemon config still exits after one cycle (existing behaviour unchanged)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -77,6 +77,7 @@ handwritten_tests = \
 	t/protocol_simplycom.pl \
 	t/protocol_spaceship.pl \
 	t/read_recap.pl \
+	t/run_once.pl \
 	t/skip.pl \
 	t/ssl-validate.pl \
 	t/update_nics.pl \
--- a/ddclient.conf.in
+++ b/ddclient.conf.in
@@ -226,7 +226,7 @@ pid=@runstatedir@/ddclient.pid  # record PID in file.
 #
 # protocol=duckdns, \
 # password=my-auto-generated-password \
-# hostwithoutduckdnsorg,host2withoutduckdnsorg
+# hostwithoutduckdnsorg
 
 ##
 ## Freemyip (http://freemyip.com/)
--- a/ddclient.in
+++ b/ddclient.in
@@ -709,6 +709,7 @@ our %cfgvars = (
             opt('daemon') ? int(opt('daemon') * 0.2) : undef
         }, undef),
         'foreground'          => setv(T_BOOL,  0, 0,                    undef),
+        'run-once'            => setv(T_BOOL,  0, 0,                    undef),
         'file'                => setv(T_FILE,  0, "$etc/$program.conf", undef),
         'cache'               => setv(T_FILE,  0, "$cachedir/$program.cache", undef),
         'pid'                 => setv(T_FILE,  0, undef,                undef),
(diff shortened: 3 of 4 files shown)


indrajitr:
Do we need another cli switch? Not specifying `--daemon` in CLI or `daemon=...` in `ddclient.conf` is effectively treated as run-once.

So essentially, `ddclient --verbose --debug` (without `--daemon` in CLI or in config) is executed only once. And `ddclient --verbose --debug --foreground` prevents forking child process but still excuted only once.

Edit: See the AI agent analysis in the section "3. One-shot / cron-style execution" https://github.com/ddclient/ddclient/issues/609#issuecomment-4537130715. That analysis is actually accurate! We need to test it against current 4.x of course, but I have extensively used it this way in the past via systemd unit `ddclient.{service,timer}`.

agross:
Working as intended!

beadon:
> Working as intended!

@agross , I have checked in with @indrajitr here on this one, and it seems that including this will introduce more "flag spam" to ddclient.  We agree that this should probably not proceed.

The main problem is with the requirement ddclient has around using daemon for all things.  This was originally a requirement to manage the state of the frequently running process, but now tools like systemd have stepped up to provide a consistent means to manage daemons, so this ddclient daemon mode is effectively obsolete.

Have you tried using :  ```ddcleint --foreground``` to perform this "one shot" instead ?

agross:
> Have you tried using : ddcleint --foreground to perform this "one shot" instead ?

Yes. From the [issue](https://github.com/ddclient/ddclient/issues/856):

> With 4.x `ddclient --daemon 0 --foreground --debug --verbose` also waits with the config-file based timeout.

Your own analysis showed that --daemon is always clamped to 60.

beadon:
Ok, yes, that is what I remembered too.  Thank you for the reminder.

Then what is called for here is a restructure of the daemon, not another flag. :-/

agross:
> Then what is called for here is a restructure of the daemon, not another flag.

It has been working fine with 3.x. It allowed `--daemon=0` and behaved like the "oneshot" (`--run-once`) mode I'm asking for.

BTW, my Fedora systemd service invokes with `--foreground`, too, but uses the daemon interval from the config file.

I think allowing `--daemon=0` to be overwritten via the CLI (again, like in 3.x) is the way to go without much restructuring required.

PR #901, opened 2026-05-25
build: install ddclient.conf.sample, never clobber live ddclient.conf

### LINKED ISSUE #823: should make install overwrite existing conf files?
My sysadmin fu is barely strong enough to be useful, and often weak enough to be damaging, so i'll ask this meekly...

I had to download the most current ddclient and install from source to work around the recent cloudflare 1.0.1.1 breakage (thank you very much for a super quick fix on that, BTW).  However, when I did `make install` the installations script(s) silently overwrote my existing ddclient.conf file with the default one that was in the source tarball.  This broke my server connectivity for a few days and hurt my feelings.

Is this by design, or by accident, or by convention?  Like most other pseudo-noobs I mostly just use whatever the distro packages are and apt-upgrade things.  BUT - I feel like I've done enough installations from source using basic makefiles that I would have run across this side effect before if it was common.

I know there's a way to avoid this (back up the conf files, more carefully read the makefile, etc.).  But it feels like MOST of the time users would prefer to NOT have their existing configs overwritten? Maybe I'm wrong and this is just how things work conventionally, but I figured it was worth asking.  I can see a few paths forward:

1) tell me to shut up and live with it, because This Is The Way.
2) have the installation scripts throw an interactive query/response
3) have the installation rename any existing .conf files to something like .conf.old or .conf.[timestamp]
3b) have the installation leave existing .conf files where they are and also copy/create the new/default files at some other location (e.g. `.ddclient.conf.default`)

4) throw a notification string during the install that says "hey I'm going to overwrite this" or "hey I just moved the file I found to <new location>"  (Seems like this might be applicable almost no matter what else we choose?)

beadon:
Closes #823

## Summary

- `make install` previously used automake's `conf_DATA` to install `ddclient.conf` directly, silently overwriting any existing config — including credentials — on every upgrade.
- This implements **Strategy 2** from #823: install a `ddclient.conf.sample` unconditionally and never touch the user's live `ddclient.conf`.
- The `ddclient.conf.sample` has been fully updated from 3.x to 4.x syntax and expanded with new provider examples so it works as an accurate reference for new users.

## Changes

**`Makefile.am`**
- Removed `conf_DATA = ddclient.conf` (the clobbering rule).
- Added a `ddclient.conf.sample: ddclient.conf` copy rule so the sample is always freshly built.
- Declared `sampledir = $(confdir)` / `sample_DATA = ddclient.conf.sample` so automake installs and uninstalls the sample correctly.
- Added `ddclient.conf.sample` to `CLEANFILES`.

**`README.md`**
- Updated Manual Installation step 3: first-time users now `cp ddclient.conf.sample ddclient.conf && $EDITOR`.
- Added a note that upgrades are safe and the sample can be diffed against the live config.

**`ddclient.conf.sample`** (new file — updated for 4.x)

*3.x → 4.x syntax fixes:*
- Replaced deprecated `use=`/`fw=`/`if=`/`cmd=` options with their 4.x equivalents: `usev4=`/`fwv4=`/`ifv4=`/`cmdv4=`.
- Updated default web IP service note: default is now `ipify-ipv4`, not `checkip.dyndns.org`.
- **CloudFlare**: removed broken inline comments after `\` continuations; split into two clear examples (Global API key and API token with `login='token'`); removed the `ttl=1` that the protocol ignores.
- **GoDaddy**: fixed duplicate `password=` field — first field is `login=` (API key), second is `password=` (API secret).
- **dnsexit**: removed defunct `protocol=dnsexit` (removed in 4.x); replaced with `protocol=dnsexit2`.
- **dslreports1**: removed incorrect `server=members.dslreports.com`; the protocol default is `www.dslreports.com`.
- **NoIP**: removed redundant `ssl=yes` (now the global default).
- **Porkbun**: merged two duplicate host lines into a single correct entry.
- **nsupdate.info IPv6**: `usev6=if, if=eth0` → `usev6=ifv6, ifv6=eth0`; removed outdated 3.8.1 note.
- **MyOnlinePortal**: replaced invalid `ipv6=yes`, deprecated `if-skip=`, and redundant `ssl=yes` with correct 4.x options.
- **Hetzner**: corrected placeholder from `api-key` to `api-token`.
- **Njalla**: `quietreply=no|yes` → `quietreply=no`; updated password placeholder.
- **Yandex**: removed stray trailing `\` from the last host line.
- **NameCheap**: clarified password placeholder is a generated DDNS password.

*New provider examples:*
- **DynDNS** (dyndns.org): expanded to a proper dual-stack example with `usev4=webv4` + `usev6=ifv6`; noted that `password=` is the updater client key, not the account password.
- **deSEC**: free DNSSEC-enabled provider (Germany); dual-stack via `checkipv4.dedyn.io` / `checkipv6.dedyn.io`; noted that the dedyn.io hostname is the Basic Auth username.
- **dynv6**: free IPv6-first provider; noted that the `dyndns2` endpoint updates IPv4 only — IPv6 requires their native API (not yet in ddclient).
- **LuaDNS**: DNS-as-code provider with dual-stack `dyndns2` support; noted requirement to pre-create A/AAAA records in the control panel.

## Behaviour after this change

| Path | Before | After |
|------|--------|-------|
| `/etc/ddclient/ddclient.conf` | Overwritten silently on every `make install` | **Never touched** |
| `/etc/ddclient/ddclient.conf.sample` | Not installed | Installed/refreshed on every `make install` |

## Packager note

RPM (`%config(noreplace)`) and Debian (`postinst` copy) packagers are unaffected — they already handle the first-install/upgrade distinction correctly at the package level. This fix closes the gap for users installing from a source tarball via `make install`.

## Test plan

- [x] `./autogen && ./configure --prefix=/usr --sysconfdir=/etc --localstatedir=/var && make` — builds cleanly, `cp` rule fires for `ddclient.conf.sample`
- [x] `make VERBOSE=1 check` — 743 pass, 6 skip (pre-existing optional-module skips), 0 fail
- [x] `make install --dry-run` — confirms `ddclient.conf.sample` goes to `/etc/ddclient/`; `ddclient.conf` does not appear in the install targets
- [x] All protocol examples in `ddclient.conf.sample` verified against `ddclient.in` source for 4.x correctness
- [x] deSEC, dynv6, and LuaDNS endpoint details verified against provider documentation

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -18,7 +18,8 @@ CLEANFILES =
 
 subst_files = ddclient ddclient.conf
 EXTRA_DIST += $(subst_files:=.in)
-CLEANFILES += $(subst_files)
+CLEANFILES += $(subst_files) ddclient.conf.sample
+EXTRA_DIST += ddclient.conf.sample
 
 $(subst_files): Makefile
 	rm -f '$@' '$@'.tmp
@@ -39,9 +40,13 @@ $(subst_files): Makefile
 ddclient: $(srcdir)/ddclient.in
 ddclient.conf: $(srcdir)/ddclient.conf.in
 
+ddclient.conf.sample: ddclient.conf
+	cp '$<' '$@'
+
 bin_SCRIPTS = ddclient
 
-conf_DATA = ddclient.conf
+sampledir = $(confdir)
+sample_DATA = ddclient.conf.sample
 
 install-data-local:
 	$(MKDIR_P) '$(DESTDIR)$(localstatedir)'/cache/ddclient
--- a/README.md
+++ b/README.md
@@ -131,7 +131,16 @@ operating system. See the image to the right for a list of distributions with a 
      sudo make install
      ```
 
-  3. Edit `/etc/ddclient/ddclient.conf`.
+  3. Copy the sample config and edit it for your setup:
+
+     ```shell
+     sudo cp /etc/ddclient/ddclient.conf.sample /etc/ddclient/ddclient.conf
+     sudo $EDITOR /etc/ddclient/ddclient.conf
+     ```
+
+     `make install` no longer overwrites an existing `ddclient.conf`, so upgrades
+     are safe.  The sample at `ddclient.conf.sample` is always refreshed on every
+     install so you can diff it against your live config to see what's new.
 
 #### systemd
 
--- a/ddclient.conf.sample
+++ b/ddclient.conf.sample
@@ -1,0 +1,457 @@
+######################################################################
+##
+## Define default global variables with lines like:
+##      var=value [, var=value]*
+## These values will be used for each following host unless overridden
+## with a local variable definition.
+##
+## Define local variables for one or more hosts with:
+##      var=value [, var=value]* host.and.domain[,host2.and.domain...]
+##
+## Lines can be continued on the following line by ending the line
+## with a \
+##
+##
+## Warning: not all supported routers or dynamic DNS services
+##          are mentioned here.
+##
+######################################################################
+
+## Use encryption (TLS) when the scheme (either "http://" or "https://") is
+## missing from a URL.  Defaults to "yes".
+#ssl=yes
+
+daemon=300                      # check every 300 seconds
+jitter=60                       # apply up to 60 seconds of delay to the daemon interval
+syslog=yes                      # log update msgs to syslog
+mail=root                       # mail all msgs to root
+mail-failure=root               # mail failed update msgs to root
+# mail-from=root                # set the email "From:" header to "root".  If
+                                # unset (the default) or empty, the from address
+                                # depends on your system's default behavior.
+pid=/usr/local/var/run/ddclient.pid  # record PID in file.
+# postscript=script             # run script after updating.  The new IP is
+                                # added as argument.
+#
+#use=watchguard-soho,        fw=192.168.111.1:80        # via Watchguard's SOHO FW
+#use=netopia-r910,           fw=192.168.111.1:80        # via Netopia R910 FW
+#use=smc-barricade,          fw=192.168.123.254:80      # via SMC's Barricade FW
+#use=netgear-rt3xx,          fw=192.168.0.1:80          # via Netgear's internet FW
+#use=linksys,                fw=192.168.1.1:80          # via Linksys's internet FW
+#use=maxgate-ugate3x00,      fw=192.168.0.1:80          # via MaxGate's UGATE-3x00  FW
+#use=elsa-lancom-dsl10,      fw=10.0.0.254:80           # via ELSA LanCom DSL/10 DSL Router
+#use=elsa-lancom-dsl10-ch01, fw=10.0.0.254:80           # via ELSA LanCom DSL/10 DSL Router
+#use=elsa-lancom-dsl10-ch02, fw=10.0.0.254:80           # via ELSA LanCom DSL/10 DSL Router
+#use=alcatel-stp,            fw=10.0.0.138:80           # via Alcatel Speed Touch Pro
+#use=xsense-aero,            fw=192.168.1.1:80          # via Xsense Aero Router
+#use=allnet-1298,            fw=192.168.1.1:80          # via AllNet 1298 DSL Router
+#use=3com-oc-remote812,      fw=192.168.0.254:80        # via 3com OfficeConnect Remote 812
+#use=e-tech,                 fw=192.168.1.1:80          # via E-tech Router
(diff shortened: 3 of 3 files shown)


indrajitr:
How is `ddclient.conf.sample` going to be different from `ddclient.conf.in`? Can we reuse `ddclient.conf.in`?

beadon:
ddclient.conf.in is not out on the installed filesystem at install time. People are not using make tools to install config files.  Additionally, since everyone is installing via a package, there is the expectation to have example files included.  The typical sys admin today is not using make files to manage configs.

I am also following in the footsteps of a number of other large applications that follow this pattern.  See linked issue.

Overwriting the existing ddclient.conf is very bad form causing people to lose working configs on a package or binary update.

indrajitr:
> ddclient.conf.in is not out on the installed filesystem at install time. People are not using make tools to install config files. Additionally, since everyone is installing via a package, there is the expectation to have example files included. The typical sys admin today is not using make files to manage configs.

1. Debian family:
   - The config `/etc/ddclient.conf` is `debconf` generated and is never over-written if it exists. This means if the file exists, you either modify it manually or go through `dpkg-reconfigure`
   - `ddclient.conf.in` is installed in `/usr/share/doc/ddclient/examples/sample-etc_ddclient.conf` for reference.

2. RedHat family:
   - `ddclient.conf.in` is installed in `/etc/ddclient.conf` but if and only if `/etc/ddclient.conf` has not been manually modified. Modification that the package manager is unaware of results in creation of `.rpmsave` or `.rpmnew` to avoid being overwritten.

> 
> I am also following in the footsteps of a number of other large applications that follow this pattern. See linked issue.
> 
> Overwriting the existing ddclient.conf is very bad form causing people to lose working configs on a package or binary update.

Summary:
1. Preserving `ddclient.conf` is indeed a must, that's not being disputed here.
2. Two major Linux packaging ecosystem _actually preserves_ `ddclient.conf.in` - either as sample doc (exactly what @beadon is alluding to) or as sample /etc/ddclient.conf which is then expected to be modified manually (or through IaaC toolchain).
3. If `make install` is overwriting existing `ddclient.conf` file, it is a serious bug, we should absolutely fix it. We can install it beside the existing `ddclient.conf` and `ddclient.conf.{sample,new,whatever}`.
4. Keeping around both seemingly similar `ddclient.conf.in` *and* `ddclient.conf.sample` in the codebase is just adding to the confusion imho. @beadon I am curious to know about your though process on why we should keep both of them and when we use one vs the other or how we articulare/document the usage of one vs the other.

beadon:
> If make install is overwriting existing ddclient.conf file, it is a serious bug, we should absolutely fix it. We can install it beside the existing ddclient.conf and ddclient.conf.{sample,new,whatever}.

Unless I've done something incorrect, it is.

>Keeping around both seemingly similar ddclient.conf.in and ddclient.conf.sample in the codebase is just adding to
> the confusion imho. @beadon I am curious to know about your though process on why we should keep both of
> them and when we use one vs the other or how we articulare/document the usage of one vs the other.

I agree, is your proposal to build the sample conf from the ddclient.conf.in using Make then, instead of having a separate sample config file ?   Since we'e established no one uses ddclient.conf.in / Makefile workflow I was moving to obsolete it.

beadon:
agreed that we need to have the sample file spit out from the makefile, it should be built from the makefile.

beadon:
Rebased onto upstream/main (a35f724). The rebase applied cleanly with no conflicts — the single commit landed on top of the latest upstream without any merge issues.

StantonMatt (commented):
I reproduced the Debian failure locally against `d8cb286` in a Debian stable container. The Perl tests themselves pass:

`make VERBOSE=1 AM_COLOR_TESTS=always check` reports `TOTAL: 967`, `PASS: 943`, `SKIP: 24`, `FAIL: 0`, `ERROR: 0`.

The red checks are coming from the distribution tarball comparison after `distcheck`. `ddclient.conf.sample` is tracked in this branch, but the build regenerates it from `ddclient.conf.in`; the regenerated file has provider examples that are not in the checked-in `ddclient.conf.sample` yet. The first mismatch I see is the missing Joker.com section, followed by Dynadot, DDNSS.de, IONOS, NS1, DynV6, Sitelutions, Dynu, and the later RFC2136 examples.

In the scratch checkout, running through configure/build leaves exactly that file dirty:

`git diff -- ddclient.conf.sample`

shows 213 inserted lines. So the likely minimal fix is to regenerate `ddclient.conf.sample` from the current `ddclient.conf.in` and commit that updated file. After that, the Debian-like jobs should get past `dist-tarball-check`; the install behavior itself is already showing the intended target:

`/usr/bin/install -c -m 644 ddclient.conf.sample .../etc/ddclient`

PR #921, opened 2026-05-27
feat: add interactive install.sh setup script

beadon:
## Summary

- Adds `install.sh`, an interactive shell script that guides a new user through the full configure/build/install flow on Linux (and macOS via Homebrew)
- Detects the OS and package manager (`apt`, `dnf`/`yum`, `pacman`, `zypper`, `brew`), checks for and installs missing dependencies via the native package manager
- Detects GNU make (`gmake` vs `make`) — important on BSD and macOS where the system `make` is not GNU make
- Handles both **git clone** builds (offers to run `autogen`, checks for `autoconf`/`automake` only when needed) and **release tarball** builds (skips autogen entirely)
- Prompts for configure options (`--prefix`, `--sysconfdir`, `--localstatedir`) with sensible defaults
- Runs `make VERBOSE=1 check` (per README recommendation), prompts to continue if tests fail
- Installs with `sudo` only for system-wide paths; post-install optionally sets up the systemd service and creates a starter config
- `--non-interactive` flag accepts all defaults for scripted/CI use
- Included in `EXTRA_DIST` so it ships in release tarballs
- Uses `--foreground` (not `--daemon=0`) in the post-install test hint — `--foreground` explicitly prevents forking regardless of the `daemon=` value in the config file

## Test plan

- [x] Syntax checked with `bash -n install.sh`
- [x] Smoke-tested with `--non-interactive` on Debian/Ubuntu: correctly detects `gmake`, skips autogen (configure present), runs full test suite, reaches sudo step
- [x] All existing `make check` tests continue to pass

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,6 +7,7 @@ EXTRA_DIST = \
 	README.cisco \
 	README.md \
 	autogen \
+	install.sh \
 	sample-ddclient-wrapper.sh \
 	sample-etc_cron.d_ddclient \
 	sample-etc_dhclient-exit-hooks \
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,399 @@
+#!/bin/bash
+# install.sh — Interactive installer for ddclient
+#
+# Detects your OS, installs missing dependencies, configures, builds, and
+# installs ddclient.  Run as a normal user; sudo is invoked only when needed.
+#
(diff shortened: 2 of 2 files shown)


beadon:
Rebased onto upstream/main (a35f724). No conflicts — the rebase applied cleanly in a single pass.

PR #958, opened 2026-06-14
Fix command part of URL for dinahosting provider

cquike:
Hi,
I have tried to use ddclient for dinahosting provider but it looks like it is currently broken.
I managed to fix it just adding a parenthesis around the is_ipv6 condition.
Before the fix this is the URL I get:

```
https://dinahosting.com/special/api.php?hostname=host&domain=domain.orgAAAA&ip=192.168.1.1
```
which caused an ERROR 2000.
And after the fix:
```
https://dinahosting.com/special/api.php?hostname=host&domain=domain.org&command=Domain_Zone_UpdateTypeA&ip=192.168.1.1
```
which is now working.

Thanks!

### DIFF
--- a/ddclient.in
+++ b/ddclient.in
@@ -8428,7 +8428,7 @@ sub nic_dinahosting_update {
         my $url = 'https://' . opt('server', $h) . opt('script', $h);
         $url .= "?hostname=$hostname";
         $url .= "&domain=$domain";
-        $url .= "&command=Domain_Zone_UpdateType" . is_ipv6($ip) ? 'AAAA' : 'A';
+        $url .= "&command=Domain_Zone_UpdateType" . (is_ipv6($ip) ? 'AAAA' : 'A');
         $url .= "&ip=$ip";
         my $reply = geturl(
             proxy => opt('proxy'),


cquike:
I guess that in previous versions of perl the order in the evaluation was different and the previous syntax was properly working whereas now, at least with perl 5.40 in Debian 13 it is not longer working.

beadon (approved):
The fix is correct, but the explanation in the comment about Perl version differences isn't quite right — this is a pre-existing operator precedence bug, not a regression in Perl 5.40.

In Perl, `.` (concatenation) has always had higher precedence than `?:` (ternary). So the original line:

```perl
$url .= "&command=Domain_Zone_UpdateType" . is_ipv6($ip) ? 'AAAA' : 'A';
```

is actually parsed as:

```perl
$url .= ("&command=Domain_Zone_UpdateType" . is_ipv6($ip)) ? 'AAAA' : 'A';
```

The concatenated string is always non-empty (always truthy), so the ternary always picks `'AAAA'` — and the entire `&command=...` prefix is consumed as the ternary's *condition*, never appended to `$url`. That's exactly what the buggy URL shows: `domain.orgAAAA` with no `&command=` parameter at all. It would also always emit `AAAA` even for IPv4.

This was broken from the start — operator precedence between `.` and `?:` hasn't changed across Perl versions. It likely just went unnoticed, possibly because older versions of the dinahosting API were more lenient about the malformed URL.

The fix (wrapping the ternary in parentheses) is exactly right. Thanks for tracking this down!

cquike:
Thanks a lot for the detailed explanation! I am myself not very versed at perl, so that's appreciated.

PR #964, opened 2026-07-29
fix: dynu protocol: add myipv6=no when usev6=disabled

inevitable-commit:
The commit [126797c](https://github.com/ddclient/ddclient/commit/126797c40dd51535e0f28450b1315e2f6a1466db) missed handling the case of only IPv4. This causes dynu to auto-detect IPv6 address of client making the request and use that IPv6 to update entry for IPv6 as well as update entry for IPv4 as requested.

This is handled by adding parameter myipv6=no when ddclient is configure as usev4 != disabled and usev6=disabled.


DustyRah:
I arrived at this independently and opened a duplicate (#965) before spotting this PR — apologies. I've since rescoped mine to drop the overlapping change and defer to this one. In the process I reproduced the bug and verified this exact fix against the live Dynu API, so here is the evidence in case it's useful for review.

**Environment:** dual-stack Ubuntu 24.04, IPv4 `172.81.61.141` / IPv6 `2602:ff23:50:228::1`, ddclient `4.0.1-rc.2` built from `main`.

One methodology note that cost me some time: Dynu's nameservers replicate over several minutes, and one of them served stale answers well after an update landed. Verifying via `dig` produced false negatives. All results below were instead read back through Dynu's v2 DNS API, which reflects the change immediately.

### Reproducing the bug

Sentinel AAAA of `2602:ff23:50:228::beef`, `usev6=disabled`, running unmodified `main`:

```
before: v4=203.0.113.9      v6=2602:ff23:50:228::beef
after:  v4=172.81.61.141    v6=2602:ff23:50:228::1     <-- AAAA overwritten
```

The AAAA record was replaced with the host's real IPv6 address, which is exactly the auto-detection you describe.

### Parameter semantics

Probing the API directly, the behaviour depends on the transport the request arrives over, which is what makes this bite only on dual-stack hosts:

| Request sends | Arrives over | A record | AAAA record |
|---|---|---|---|
| `myip=<v4>` | IPv4 | set | untouched |
| `myip=<v4>` | IPv6 | set | **overwritten** |
| `myipv6=<v6>` | IPv4 | **overwritten** | set |
| `myipv6=<v6>` | IPv6 | untouched | set |
| `myip=no&myipv6=<v6>` | either | untouched | set |
| `myip=<v4>&myipv6=no` | either | set | untouched |

So the existing `myip=no` guard and the `myipv6=no` guard this PR adds are both load-bearing, and only the explicit `no` makes the outcome independent of transport.

### Verifying the fix

Applying the same one-line change as this PR, same host and same sentinel:

```
before: v4=203.0.113.9      v6=2602:ff23:50:228::beef
after:  v4=172.81.61.141    v6=2602:ff23:50:228::beef  <-- preserved
```

I also confirmed the behaviour your added code comment describes: a dual-stack update sets both records but the response echoes only one address (`good 198.51.100.7` for `myip=198.51.100.7&myipv6=2602:ff23:50:228::7`, with both records updated). Treating that single acknowledgement as success for both is correct.

Two smaller findings from the same testing, in case they're ever useful: `myip=` accepts an IPv4 address only — an IPv6 address placed there alone is discarded and the server falls back to auto-detection — and the combined form is order-sensitive, with `myip=<v4>,<v6>` accepted but `myip=<v6>,<v4>` rejected.

PR #965, opened 2026-08-06
dynu: add dynu-ipv6 and dynu-ipv4 web IP check services

DustyRah:
### Problem

`checkip.dynu.com` publishes an **A record only**, so it can never return an IPv6 address. The only built-in Dynu discovery service is `dynu`, which points there — so there is no built-in way to obtain an IPv6 address for `usev6`, even though the `dynu` protocol supports IPv6 updates.

### Change

* `dynu-ipv6` → `https://checkipv6.dynu.com/` (publishes AAAA, reports the caller's IPv6 address)
* `dynu-ipv4` → `https://checkip.dynu.com/` (explicit name for the existing service)
* `dynu` is kept as a synonym for `dynu-ipv4`, so existing configurations are unaffected.

This matches the `-ipv4`/`-ipv6` naming already used for `noip`, `desec`, `joker`, `nsupdate.info` and `ipify`.

Both endpoints were checked against the live services from a dual-stack Ubuntu 24.04 host. `checkip.dynu.com` returned the host's IPv4 address and `checkipv6.dynu.com` returned its IPv6 address, both in the `Current IP Address: <address>` form ddclient already parses:

```
$ curl -s https://checkip.dynu.com/
Current IP Address: 172.81.61.141
$ curl -s https://checkipv6.dynu.com/
Current IP Address: 2602:ff23:50:228::1
```

DNS confirms the asymmetry that motivates this:

```
checkip.dynu.com     A only,    no AAAA
checkipv6.dynu.com   AAAA only, no A
```

Also documents the IPv6-only and dual-stack configurations in `nic_dynu_examples` and `ddclient.conf.in`, which the protocol supported but the examples did not show.

### Relationship to #964

This PR **does not** touch `nic_dynu_update`. An earlier version of it also added the missing `myipv6=no` guard for IPv4-only updates, but #964 by @inevitable-commit already fixes exactly that and predates this PR, so that change has been dropped here in favour of #964.

The two are complementary and independent — #964 fixes the update logic, this adds the missing IPv6 discovery service — and they touch different parts of the file, so they should not conflict in either merge order.

### Tests

No test changes: this adds entries to `%builtinweb` and documentation only. Full suite on Ubuntu 24.04 / Perl 5.38 with `make check`: **965 pass, 0 fail, 0 error** (6 skips, all `HTTP::Daemon::SSL not available` in the local environment).


DustyRah:
Rescoped this PR. It originally also added the `myipv6=no` guard for IPv4-only updates, but I hadn't spotted that #964 already fixes exactly that and predates this PR. I've dropped that change here and rebased onto `main` — the fix belongs to #964.

What remains is independent of #964 and doesn't touch `nic_dynu_update`: the `dynu-ipv6` discovery service (`checkip.dynu.com` is IPv4-only, so there was no built-in way to get an IPv6 address for `usev6`), `dynu-ipv4` as an explicit name, and the IPv6-only/dual-stack examples. Apologies for the duplication.

